PR #64, opened 2020-09-15
Add new heat source

stvdwtt:
This adds a new class, HeatSource, that should be a drop-in replacement for ElectronBeam. It reads in a scan path segment file using the syntax that is used for OpenFOAM-AM and Thesis at the MDF. It also uses the Goldak source type used by OpenFOAM-AM.

Currently this is disconnected from the rest of the code so no calculations can actually be run with it.


stvdwtt:
Per the conversation earlier, I'll split this into two PRs.

PR #65, opened 2020-09-15
New scan path parser and calculator

stvdwtt:
This reads in a scan path segment file using the syntax that is used for OpenFOAM-AM and Thesis at the MDF. Until the new heat source code is added this is free-floating in the code and cannot be used for actual simulations.


source/ElectronBeam.cc
```diff
@@ -180,22 +180,33 @@ double ElectronBeam<dim>::value(dealii::Point<dim> const &point,
...
+    /*
+    double const four_ln_pone = 4. * std::log(0.1);
     double heat_source = 0.;
     heat_source =
         -_beam.energy_conversion_eff * _beam.control_eff * _beam.max_power *
         four_ln_pone /
         (dealii::numbers::PI * _beam.diameter_squared * _beam.depth) *
         std::exp(four_ln_pone * xpy_squared / _beam.diameter_squared) *
         distribution_z;
+    */
```
Rombur, line 201: Just remove the commented code.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
+/* Copyright (c) 2016 - 2020, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
+
+#ifndef SCAN_PATH_TEMPLATES_HH
+#define SCAN_PATH_TEMPLATES_HH
```
Rombur, line 9: You don't need the `ifndef` in the `.cc` I know it exists in a couple of files but there are left over from a code refactor. They shouldn't be there.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+
+#include <ScanPath.hh>
+#include <instantiation.hh>
```
Rombur, line 12: You don't need this. This file only contains macros that can be used for explicit instantiations.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+#include <utils.hh>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+
+#include <cstdlib>
+
+using std::pow;
```
Rombur, line 20: You don't need this.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+  // Parse the scan path
+  ASSERT_THROW(boost::filesystem::exists(scan_path_file),
+               "The file " + scan_path_file + " does not exist.");
+  std::ifstream file;
+  file.open(scan_path_file);
+  std::string line;
+  int line_index = 0;
+  while (getline(file, line))
+  {
+    std::cout << line << std::endl;
```
Rombur, line 40: Leftover from debugging?

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+
+      // Set the segment type
+      ScanPathSegmentType segment_type;
+      if (split_line[0] == "0")
+      {
+        if (_segment_list.size() == 0)
+        {
+          std::string message =
+              "Error: Scan paths must begin with a 'point' segment.";
+          throw std::runtime_error(message);
```
Rombur, line 58: Just use ```ASSERT_THROW(false, "Error: Scan paths must begin with a 'point' segment.");```

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+      }
+      else if (split_line[0] == "1")
+      {
+        segment_type = ScanPathSegmentType::point;
+      }
+      else
+      {
+        std::string message = "Error: Mode type in scan path file line " +
+                              std::to_string(line_index) + "not recognized.";
+        throw std::runtime_error(message);
```
Rombur, line 70: Same comment as above.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+      }
+      _segment_list.push_back(segment);
+    }
+    line_index++;
+  }
+  file.close();
+}
+
+void ScanPath::update_current_segment_info(
+    double time, dealii::Point<2> &segment_start_point,
```
Rombur, line 109: Why is the `Point` 2D? Should it be 2D or 3D depending on the dimensionality of the domain?

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+    double &segment_start_time) const
+{
+  // Get to the correct segment (assumes that the current time is never before
+  // the current segment starts)
+  if (time > _segment_list[_current_segment].end_time)
```
Rombur, line 114: I don't think the `if` is necessary here. If the condition is false, you won't get in the `while` loop anyway.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+  }
+  else
+  {
+    segment_start_time = 0.0;
+    segment_start_point = _segment_list[_current_segment].end_point;
+  }
+}
+
+double ScanPath::value(dealii::Point<1> const &time,
+                       unsigned int const component) const
```
Rombur, line 135: Is there a point in deriving from `dealii::Function` anymore? You force `component == 1` and it is strange to store the `time` in a `dealii::Point`.

source/ScanPath.cc
```diff
@@ -0,0 +1,190 @@
...
+{
+  // The global coordinate system is (x,z,y), while the scan path coordinate
+  // system is (x,y,z). I need to convert the global coordinate number to the
+  // scan path coordinates.
```
Rombur, line 139: I am not sure what you mean here? What's the difference between the global coordinate and the scan path coordinate?

source/ScanPath.hh
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
+
+#ifndef SCAN_PATH_HH
+#define SCAN_PATH_HH
+
+#include <utils.hh>
```
Rombur, line 11: You probably don't need this.

source/ScanPath.hh
```diff
@@ -0,0 +1,132 @@
...
+ * end point. Everything one needs can be determined from these three quantities
+ * (and the segment info from the preceding segment) but in the future it might
+ * be worth adding in some redundant information like start time/point and
+ * velocity.
+ */
+struct ScanPathSegment
+{
+  double end_time;            // Unit: seconds
+  double power_modifier;      // Dimensionless
+  dealii::Point<2> end_point; // Unit: m (NOTE: converted from mm in the file)
```
Rombur, line 48: Why is this only a 2D point? Shouldn't this be a 3D point? I guess it is fine for now but we won't we need to change this when we have multiple layer?

tests/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ list(APPEND
      test_thermal_physics
      test_timer
      test_utils
+     test_scan_path
```
Rombur, line 21: Can you move this up so that it is in alphabetical order.

source/ScanPath.hh
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Returns the power coefficient for the current segment
+   */
+  double get_power_modifier(dealii::Point<1> const &time) const;
+
+  /**
+   * Method to save the segment number as a specific time. (This is currently
+   * unused in the code).
+   */
+  void save_time();
```
Rombur, line 89: Yeah I don't think this is necessary. The `save_time` and `rewind_time` are needed for some embedded time stepping methods but there is something in deal.II that takes care of that now.

source/ScanPath.hh
```diff
@@ -0,0 +1,132 @@
...
+private:
+  /**
+   * The list of information about each segment in the scan path.
+   */
+  std::vector<ScanPathSegment> _segment_list;
+
+  /**
+   * The index of the current segment in the scan path.
+   */
+  mutable unsigned int _current_segment;
```
Rombur, line 106: If we don't derive from `dealii::Function` we don't need this be `mutable` anymore.

source/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set(Adamantine_HEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/Timer.hh
     ${CMAKE_CURRENT_SOURCE_DIR}/utils.hh
     ${CMAKE_CURRENT_SOURCE_DIR}/types.hh
+    ${CMAKE_CURRENT_SOURCE_DIR}/ScanPath.hh
```
Rombur, line 16: Can you move this up to before `ThermalOperator.hh`

source/CMakeLists.txt
```diff
@@ -24,6 +25,7 @@ set(Adamantine_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/ThermalOperator.cc
     ${CMAKE_CURRENT_SOURCE_DIR}/ThermalPhysics.cc
     ${CMAKE_CURRENT_SOURCE_DIR}/Timer.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/ScanPath.cc
```
Rombur, line 28: Can you move this up before `ThermalOperator.cc`

Rombur (changes requested):
This looks pretty good. I have a few comments but most of them are cosmetic. There are two things we need to talk about:
1. Should `ScanPath` derive from `dealii::Function`? A big part of the reason `ElectronBeam` derive from `dealii::Function` is because I wanted to use `dealii::FunctionParser`.
2. Shouldn't we use 3D points for the segments instead of 2D points? Here I am mainly thinking about the case where we have multiple layers.

stvdwtt:
Thanks for the detailed review @Rombur. I'll make those changes by our next meeting on Tuesday.

Regarding the two issues in your top comment:
1. Not having ScanPath derive from dealii::Function sounds good to me. At one point I had it that way, but then changed it back in case I was missing something about the reason PointSource derived from it -- but it makes sense that it was for the function parser. I agree that it is cleaner without deriving from dealii::Function.
2. I also waffled back and forth on this. I was thinking that we might want to handle the height of the heat source somewhere else, since I think we want/need it to be on the face of the top-most activated cells. I'm not sure how we want to handle interactions with the mesh -- if we want to change the mesh to match the layer heights or the layer heights to match the mesh. Regardless, I can change the segment ends to 3D points and we can add layers of checks (e.g. that the height monotonically increases) and other processing steps (e.g. determining when a new layer starts, shifting the layer height if need be, etc) later on. How does that sound?

Rombur:
Sounds good. Thanks.

source/ScanPath.cc
stvdwtt: Correct, I'm deleting it.

source/ScanPath.cc
stvdwtt: You are correct, that must be left over from an earlier version -- I'm taking it out.

source/ScanPath.cc
stvdwtt: I think I got it slightly wrong in this comment. The issue is that the z direction in the Goldak heat source model (and the z direction in the EB heat source model in ElectronBeam) is what I'd normally call the y direction in the FE mesh (i.e. index 1 out of 0-2). The two sections of code have the last two indices flipped with respect to one another. This isn't a problem, but it just means that some of the bookkeeping in this section looks a little weird.

I'll clarify the comment.

source/ScanPath.cc
stvdwtt: I'm going to make the points in the scan path always 3D. This makes it a direct representation of the scan path in the file (which is 3D). It also leaves the possibility of having a 2D simulation where the isn't directly over the slice in question (the physics of this are a bit iffy, but that's the case for these 2D simulations in general.

source/ScanPath.cc
stvdwtt: Per the comment below, I'm removing the dealii::Function parent.

source/ScanPath.cc
stvdwtt: Actually, I'm going to kick this to the new HeatSource class. That's where the coordinate rotation should happen.

source/ScanPath.hh
stvdwtt: Agreed. See comment above.

stvdwtt:
@Rombur I made the changes we discussed, let me know if you see anything else that should be changed before this can be merged.

source/ScanPath.cc
```diff
@@ -0,0 +1,151 @@
...
+
+      // Set the segment type
+      ScanPathSegmentType segment_type;
+      if (split_line[0] == "0")
+      {
+        if (_segment_list.size() == 0)
+        {
+          ASSERT_THROW(false,
+                       "Error: Scan paths must begin with a 'point' segment.");
+        }
```
Rombur, line 47: You could just write 
```
 ASSERT_THROW(_segment_list.size() == 0, "Error: Scan paths must begin with a 'point' segment.");
```
and get ride of the `if`

source/ScanPath.cc
```diff
@@ -0,0 +1,151 @@
...
+        segment_type = ScanPathSegmentType::line;
```
Rombur, line 48: I am a little bit confused, the error message says that we need to start with a point but then you set the `segment_type` to a line.

Rombur (commented):
This looks good. I just have one comment and one question left.

source/ScanPath.cc
stvdwtt: Yes, definitely cleaner that way.

source/ScanPath.cc
stvdwtt: The assertion is to make sure that the segment in question isn't the first segment in the list. If it is the first segment, then having it marked as a line in the input file is a problem (violates the spec) and an exception is thrown. If it isn't the first segment, then it's fine that it is a line segment and the ```segment_type``` is set to point. 

Does that make sense? I added a comment to clarify in the code.

Rombur (approved):

PR #66, opened 2020-09-23
New HeatSource class (and electron beam/Goldak derived classes)

stvdwtt:
This is still free-floating in the code. My next PR will be to replace the use of ElectronBeam with this.

### DIFF
--- a/source/ElectronBeamHeatSource.cc
+++ b/source/ElectronBeamHeatSource.cc
@@ -32,17 +32,17 @@ double ElectronBeamHeatSource<dim>::value(dealii::Point<dim> const &point,
   else
   {
     double const distribution_z =
-        -3. * pow(z / HeatSource<dim>::_beam.depth, 2) -
+        -3. * std::pow(z / HeatSource<dim>::_beam.depth, 2) -
         2. * (z / HeatSource<dim>::_beam.depth) + 1.;
 
     dealii::Point<3> const beam_center =
         HeatSource<dim>::_scan_path.value(time);
-    double xpy_squared = pow(point[0] - beam_center(0), 2);
+    double xpy_squared = std::pow(point[0] - beam_center[0], 2);
     if (dim == 3)
     {
       // NOTE: Due to the differing coordinate systems, "y" here is the third
       // component of the input point.
-      xpy_squared += pow(point[2] - beam_center(1), 2);
+      xpy_squared += std::pow(point[2] - beam_center[1], 2);
     }
     double segment_power_modifier =
         HeatSource<dim>::_scan_path.get_power_modifier(time);
--- a/source/ElectronBeamHeatSource.cc
+++ b/source/ElectronBeamHeatSource.cc
@@ -1,0 +1,66 @@
+/* Copyright (c) 2016 - 2020, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
+
+#include <ElectronBeamHeatSource.hh>
+#include <instantiation.hh>
+
+namespace adamantine
+{
+
+template <int dim>
+ElectronBeamHeatSource<dim>::ElectronBeamHeatSource(
+    boost::property_tree::ptree const &database)
+    : HeatSource<dim>(database)
+{
+}
+
+template <int dim>
+double ElectronBeamHeatSource<dim>::value(dealii::Point<dim> const &point,
+                                          double const time)
+{
+  // NOTE: Due to the differing coordinate systems, "z" here is the second
+  // component of the input point.
+  double const z = point[1] - HeatSource<dim>::_max_height;
+  if ((z + HeatSource<dim>::_beam.depth) < 0.)
+  {
+    return 0.;
+  }
+  else
+  {
+    double const distribution_z =
+        -3. * pow(z / HeatSource<dim>::_beam.depth, 2) -
+        2. * (z / HeatSource<dim>::_beam.depth) + 1.;
+
+    dealii::Point<3> const beam_center =
+        HeatSource<dim>::_scan_path.value(time);
+    double xpy_squared = pow(point[0] - beam_center(0), 2);
+    if (dim == 3)
+    {
+      // NOTE: Due to the differing coordinate systems, "y" here is the third
+      // component of the input point.
+      xpy_squared += pow(point[2] - beam_center(1), 2);
+    }
+    double segment_power_modifier =
+        HeatSource<dim>::_scan_path.get_power_modifier(time);
+
+    // Electron beam heat source equation
+    double heat_source =
+        -HeatSource<dim>::_beam.absorption_efficiency *
+        HeatSource<dim>::_beam.max_power * segment_power_modifier *
+        (std::log(0.1)) /
+        (dealii::numbers::PI * HeatSource<dim>::_beam.radius_squared *
+         HeatSource<dim>::_beam.depth) *
+        std::exp((std::log(0.1)) * xpy_squared /
+                 HeatSource<dim>::_beam.radius_squared) *
+        distribution_z;
+
+    return heat_source;
+  }
+}
+} // namespace adamantine
+
+INSTANTIATE_DIM(ElectronBeamHeatSource)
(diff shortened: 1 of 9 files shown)


source/ElectronBeamHeatSource.cc
```diff
@@ -0,0 +1,66 @@
...
+  // component of the input point.
+  double const z = point[1] - HeatSource<dim>::_max_height;
+  if ((z + HeatSource<dim>::_beam.depth) < 0.)
+  {
+    return 0.;
+  }
+  else
+  {
+    double const distribution_z =
+        -3. * pow(z / HeatSource<dim>::_beam.depth, 2) -
```
Rombur, line 35: Use `std::pow`. This `pow` is the C version. In `ElectronBeam`, there is a `using std::pow` but it doesn't really simplifies the code.

source/ElectronBeamHeatSource.cc
```diff
@@ -0,0 +1,66 @@
...
+        2. * (z / HeatSource<dim>::_beam.depth) + 1.;
+
+    dealii::Point<3> const beam_center =
+        HeatSource<dim>::_scan_path.value(time);
+    double xpy_squared = pow(point[0] - beam_center(0), 2);
```
Rombur, line 40:
```suggestion
    double xpy_squared = std::pow(point[0] - beam_center[0], 2);
```

source/ElectronBeamHeatSource.cc
```diff
@@ -0,0 +1,66 @@
...
+{
+}
+
+template <int dim>
+double ElectronBeamHeatSource<dim>::value(dealii::Point<dim> const &point,
+                                          double const time)
+{
+  // NOTE: Due to the differing coordinate systems, "z" here is the second
+  // component of the input point.
+  double const z = point[1] - HeatSource<dim>::_max_height;
```
Rombur, line 27: Now that I see this thing again, it is really strange. What about we do
```c++
double const z = point[dim-1] - HeatSource<dim>::_max_height;
```

tests/test_heat_source.cc
```diff
@@ -0,0 +1,164 @@
...
+  eb_heat_source.set_max_height(0.2);
+  eb_value = eb_heat_source.value(point2, 5.0e-7);
+  BOOST_CHECK_CLOSE(eb_value, 0.0, tolerance);
+
+  // Check the beam center 0.001 s into the second segment
+  std::cout << "Checking point 3..." << std::endl;
+  dealii::Point<2> point3(8.0e-4, 0.2);
+  goldak_heat_source.set_max_height(0.2);
+  g_value = goldak_heat_source.value(point3, 0.001001);
+  double pi_over_3_to_1p5 = pow(dealii::numbers::PI / 3.0, 1.5);
```
Rombur, line 61:
```suggestion
  double pi_over_3_to_1p5 = std::pow(dealii::numbers::PI / 3.0, 1.5);
```

source/ElectronBeamHeatSource.cc
```diff
@@ -0,0 +1,66 @@
...
+        2. * (z / HeatSource<dim>::_beam.depth) + 1.;
+
+    dealii::Point<3> const beam_center =
+        HeatSource<dim>::_scan_path.value(time);
+    double xpy_squared = pow(point[0] - beam_center(0), 2);
+    if (dim == 3)
+    {
+      // NOTE: Due to the differing coordinate systems, "y" here is the third
+      // component of the input point.
+      xpy_squared += pow(point[2] - beam_center(1), 2);
```
Rombur, line 45:
```suggestion
      xpy_squared += std::pow(point[2] - beam_center[1], 2);
```

source/GoldakHeatSource.cc
```diff
@@ -0,0 +1,61 @@
...
+  double const z = point[1] - HeatSource<dim>::_max_height;
+  if ((z + HeatSource<dim>::_beam.depth) < 0.)
+  {
+    return 0.;
+  }
+  else
+  {
+    dealii::Point<3> const beam_center =
+        HeatSource<dim>::_scan_path.value(time);
+    double xpy_squared = pow(point[0] - beam_center(0), 2);
```
Rombur, line 36:
```suggestion
    double xpy_squared = std::pow(point[0] - beam_center[0], 2);
```

source/GoldakHeatSource.cc
```diff
@@ -0,0 +1,61 @@
...
+    if (dim == 3)
+    {
+      // NOTE: Due to the differing coordinate systems, "y" here is the third
+      // component of the input point.
+      xpy_squared += pow(point[2] - beam_center(1), 2);
```
Rombur, line 41:
```suggestion
      xpy_squared += pow(point[2] - std::beam_center[1], 2);
```

source/HeatSource.cc
```diff
@@ -0,0 +1,27 @@
...
+{
+
+template <int dim>
+HeatSource<dim>::HeatSource(boost::property_tree::ptree const &database)
+    : _max_height(0.), _scan_path(database.get<std::string>("input_file"))
+{
+  // Set the properties of the heat source.
+  _beam.depth = database.get<double>("depth");
+  _beam.absorption_efficiency = database.get<double>("absorption_efficiency");
+  _beam.radius_squared = pow(database.get("diameter", 2e-3) / 2.0, 2);
```
Rombur, line 21:
```suggestion
  _beam.radius_squared = std::pow(database.get("diameter", 2e-3) / 2.0, 2);
```

tests/test_heat_source.cc
```diff
@@ -0,0 +1,164 @@
...
+  expected_value = -0.1 * 10. * 1.0 * std::log(0.1) /
+                   (dealii::numbers::PI * 0.5 * 0.5 * 0.1) *
+                   std::exp(std::log(0.1) * 1.0e-4 * 1.0e-4 / 0.25) *
+                   (-3.0 * 0.01 * 0.01 / 0.1 / 0.1 + 2.0 * 0.01 / 0.1 + 1.0);
+  BOOST_CHECK_CLOSE(eb_value, expected_value, tolerance);
+}
+
+BOOST_AUTO_TEST_CASE(heat_source_value_3d)
+{
+  double const tolerance = 1e-12;
```
Rombur, line 93: Any reason the other tolerance was `1e-10`?

Rombur (approved):
This looks good. I guess the only question is: should we fix the mess I did by reordering x,y,z

tests/test_heat_source.cc
stvdwtt: No good reason :) -- I'm switching that to 1e-12.

source/ElectronBeamHeatSource.cc
stvdwtt: Got it -- I didn't realize there was a separate C `pow`. Switching them over.

Rombur: Most math functions have a separate C version (see [here](https://en.cppreference.com/w/c/numeric/math/pow) for `pow`). The main difference is that they usually work for only one type and silently cast to it if necessary. This can produce strange bug. That's why I avoid to use them.

stvdwtt:
@Rombur As far as the coordinate system goes -- I guess it's a question of whether it is more confusing as it is now or to change the meaning of the second coordinate between 2D and 3D. Most production simulations will be in 3D, so I'd lean in the direction of changing the coordinate system so the last coordinate is the thickness of the part. Srdjan might have an opinion on this -- I'm guessing that the other codes have z for the thickness, but I don't know that.

source/ElectronBeamHeatSource.cc
stvdwtt: Good to know, thanks!

Rombur:
Yeah let's just change it. I think it makes more sense and 1. it doesn't matter for the users who don't see this anyway 2. this change is limited to the heat source classes, it won't affect the rest of the code.

stvdwtt:
I'm confused now -- wouldn't changing this mean we have to make changes in `Geometry`?

Rombur:
That's true. I forgot about it.

stvdwtt:
What do you think is the best path forward? Maybe, merge this in without changing the coordinate systems, do a PR where we replace `ElectronBeam` with a `HeatSource` child, and then a third PR changing the coordinate system in `HeatSource` and `Geometry`?

Rombur:
Sounds good but you haven't changed the `pow` or the tolerance. Did you forget to push?

stvdwtt:
Oh, you're right, I hadn't pushed yet. I just did, and I'll merge once the checks pass. Then I'll get to work on replacing `ElectronBeam`.

PR #67, opened 2020-09-29
Actually use the HeatSource classes

stvdwtt:
Given that we're almost at the end of the FY and I'll be missing the meetings on 9/29, I'm putting this up, even though it's still in progress.

This PR removes the ElectronBeam and switches over to the new HeatSource classes.

Current status:
- All tests pass on the CPU

To-dos:
- Check on GPU
- Currently the host template instantiations for ThermalPhysics are brute forced, like they are for Cuda because I haven't figured out the template instantiation macros yet.
- Allow switching between GoldakHeatSource and ElectronBeamHeat source from the input file -- currently the `run` calls in adamantine.cc are hard-coded for the GoldakHeatSource.
- Another pass for code clean up, making sure the old ElectronBeam input parameters are removed, and making sure the comments are up to date.


stvdwtt:
@Rombur, as I look at this, I think I remember that you said that `thermal_physics` should have a vector of HeatSource objects, not the derived classes so that we could in principle mix multiple types of heat sources. It would also save us from a lot of the templates on the heat source type that I had to add. I haven't worked much with base class pointers before -- as long as everything is virtual, there shouldn't be any issues, right?

Rombur:
Right. Do you want to change this yourself or do you want me to do it? That's a lot of templates right now...

stvdwtt:
I can change it (although depending on how some of my other work goes, I might not get to it until Monday).

Rombur:
Looks like you removed files in `test/data` but they are still in `tests/CMakeLists.txt` and cmake is trying (and failing) to copy them

stvdwtt:
Thanks for seeing that -- it's fixed now.

application/adamantine.hh
```diff
@@ -414,19 +409,24 @@ compute_cells_to_refine(
...
+      {
+        if (beam->value(cell->center(), current_time) > 1e-15)
+        {
           cells_to_refine.push_back(cell);
+        }
+      }
     }
   }
 
+  /* Remove?
```
Rombur, line 426: Yeah remove this and removed the functions in the `HeatSource`

source/HeatSource.hh
```diff
@@ -69,11 +69,16 @@ public:
...
-   *   - <B>input_file</B>: name of the file that contains the scan path
+   *   - <B>scan_path_file</B>: name of the file that contains the scan path
    *     segments
    */
   HeatSource(boost::property_tree::ptree const &database);
 
+  /**
+   * Destructor.
+   */
+  virtual ~HeatSource(){};
```
Rombur, line 80:
```suggestion
  virtual ~HeatSource() = default;
```

source/instantiation.hh
```diff
@@ -68,4 +68,36 @@ BOOST_PP_TUPLE_ELEM(2, TUPLE_2), dealii::MemorySpace::CUDA, quadrature>;
...
+  template class adamantine::BOOST_PP_TUPLE_ELEM(0, TUPLE_2)<BOOST_PP_TUPLE_ELEM(1, TUPLE_2),\
+BOOST_PP_TUPLE_ELEM(2, TUPLE_2), adamantine::GoldakHeatSource<2>, dealii::MemorySpace::Host, quadrature>;
+#define MA_FE_DEGREE2(z, fe_degree, TUPLE_1) \
+  BOOST_PP_SEQ_FOR_EACH(MA_QUADRATURE_TYPE2, \
+      BOOST_PP_TUPLE_REPLACE(TUPLE_1, 2, fe_degree), QUADRATURE_TYPE)
+#define MA_DIM2(z, dim, TUPLE_0) \
+  BOOST_PP_REPEAT_FROM_TO(1, 11, MA_FE_DEGREE2, BOOST_PP_TUPLE_REPLACE(TUPLE_0, 1, dim))
+#define INSTANTIATE_DIM_FEDEGREE_HS_QUAD_HOST(TUPLE_0) BOOST_PP_REPEAT_FROM_TO(2, 4, MA_DIM2, TUPLE_0)
+
+// #ifdef ADAMANTINE_HAVE_CUDA
```
Rombur, line 90: You can remove this. Unfortunately `nvcc` doesn't play well with variadic macros

stvdwtt: Can I remove the CUDA macros higher up in the file as well? It doesn't seem that they're used.

Rombur: Yes please.

Rombur (approved):

PR #68, opened 2020-11-16
Element activation

Rombur:
This PR does the following:
 - Update the code to work with a newer version of deal.II: once this is merged everybody will need to update deal.II.
 - Update the build system: in particular require a newer version of CMake and bump the C++ standard to C++17. This will allow us to use the parallel capabilities of the STL library.
 - Use hp::FECollection everywhere this is needed for element activation.

### DIFF
--- a/source/ThermalPhysics.templates.hh
+++ b/source/ThermalPhysics.templates.hh
@@ -15,8 +15,11 @@
 #include <ThermalPhysics.hh>
 
 #include <deal.II/dofs/dof_tools.h>
-#include <deal.II/fe/fe_values.h>
+#include <deal.II/fe/fe_nothing.h>
+#include <deal.II/fe/fe_q.h>
 #include <deal.II/grid/filtered_iterator.h>
+#include <deal.II/hp/fe_values.h>
+#include <deal.II/hp/q_collection.h>
 #include <deal.II/lac/precondition.h>
 #include <deal.II/lac/solver_gmres.h>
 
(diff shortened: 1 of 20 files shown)


stvdwtt:
I'll try to review this by the end of the day and before tomorrow's meeting at the latest.

source/ThermalPhysics.templates.hh
```diff
@@ -52,28 +55,12 @@ template <int dim, int fe_degree, typename MemorySpaceType,
...
-                                dealii::IteratorFilters::LocallyOwnedCell()))
-  {
-    fe_values.reinit(cell);
-    cell->get_dof_indices(local_dof_indices);
-    for (auto const dof_index : local_dof_indices)
-    {
-      if (local_elements.is_element(dof_index) == true)
-        vector[dof_index] = value;
-    }
-  }
```
stvdwtt, line 76: I still need to dive into the deal.II 9.3 docs, but I take it that this 20 lines of code is now done internally by the `local_element` function?

Rombur: The previous code could have been extended to impose a non-uniform initial temperature. The current code set just all the entries to `value` instead of going through the mesh and set entries associated to this cell to `value`. It's a more blunt approach but it's faster if we need know that all the entries in the vector will be the same. So this does not work with every type of finite elements but we are only using Lagrange finite elements.

stvdwtt (approved):

PR #70, opened 2020-11-23
Refactor heat source

Rombur:
This PR refactor the `HeatSource` to support a cube heat source. This is used to check that we don't leak energy over time. I think that's the kind of sanity check we want in the test suite.


stvdwtt (commented):
Other than the cube heat source for testing, do we expect to have other heat sources that aren't associated with a beam? If all of the "real" heat sources share some attributes like the beam quantities, it seems cleaner to include those in the base class and give the test heat source(s) dummy values rather than having to separately add them to each of the "real" heat sources classes.

With just two "real" heat sources, it isn't a big deal to have a minimalist base class, and the cube heat source is 1/3 of the cases. I suppose if we find that we need to add many more heat source classes (e.g. for wire or DED) that we could then look for commonalities and then move those to the base class.

stvdwtt (commented):
And thanks for catching that `_current_time` is no longer needed.

Rombur:
This is ready for a new review

stvdwtt (approved):
Looks good to me

PR #71, opened 2020-11-24
Fix example input file

Rombur:


stvdwtt (commented):
Thanks for catching this.

Should `application/CMakeLists.txt` also copy `input_scan_path.txt` to the `bin` directory like we do for `input.info`?

Rombur:
Yeah we should. Good point

Rombur:
This is ready to be merged. I do have a question though, do we describe the entries of `scan_path.txt` I was sure how to create a new scanning pattern.

stvdwtt (approved):

stvdwtt:
After the three-line header, here are the column descriptions:
1. Mode: 1 for "spot" mode, "0" for line mode
2.-4. (x,y,z) coordinates in units of mm. For "line" mode this is the _ending_ position of the the line.
5. The coefficient for the nominal power. Usually this is either 0 or 1, but sometimes intermediate values are used when turning a corner.
6.  In "spot" mode, this is the dwell time in seconds, in "line" mode this is the velocity in m/s.

The first entry must be a "spot". If it was a line, there would be no way to know where the line starts (since the coordinates are the ending coordinates). By convention, Jamie said that he and Alex avoid using a zero second dwell time for the first spot and instead choose some small positive number.

Does that make sense?

Rombur:
Makes sense. Thanks!

PR #73, opened 2020-12-08
Recompute the max height of the object during the simulation

Rombur:
This allows the heat sources to adapt to the addition of material. This is PR uses a single new height. This is sufficient for now and I am not sure that we are interested in cases where the new height is different in different part of the domain. That would mean that we start building part of the object and then go back down to finish another part.


stvdwtt (approved):
We might choose to do somethings differently in terms of what's in ThermalPhysics and what's in Geometry later (see #74), but for now this is good.

PR #75, opened 2020-12-14
Update README.md

Rombur:
This PR updates how to install and run the code. I've also added comments every time we get a value from a property tree. This should make it easier to maintain the README

PR #76, opened 2021-01-14
Add an input parameter for the output frequency

stvdwtt:
Simple change so that it doesn't output every time step. I'll wait until #75 is merged in and add the README text (hence the WIP).


stvdwtt:
@Rombur, I've merged in the updates from #75 and added README text, so this is now ready for your review.

Rombur (approved):

PR #78, opened 2021-01-22
Create map between x, y, and z axes and 0, 1, and 2 indices

Rombur:
The element activation code has many moving parts so instead of having one big PR, I will create a bunch of small ones. This PR fixes the problem where the y and z are switched in 3D


stvdwtt (approved):
That's a really nice way to resolve the issue, I'll remember that for future projects.

PR #79, opened 2021-01-28
Only fill part of the domain with material

Rombur:


source/ThermalPhysics.templates.hh
```diff
@@ -315,6 +315,19 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::ThermalPhysics(
...
   }
+
+  // Set material on part of the domain
+  // PropertyTreeInput geometry.material_height
+  double const material_height = database.get("geometry.material_height", 1e9);
+  for (auto const &cell : _dof_handler.active_cell_iterators())
+  {
+    // If the center of the cell is below material_height, it contains material
+    // otherwise it does not.
+    if (cell->center()[1] < material_height)
```
stvdwtt, line 326: This should be using the new enum for the axes, right? In 3D, the relevant axis index is 2, I think.

Rombur: good catch

Rombur:
I have updated the code.

stvdwtt (approved):

PR #81, opened 2021-02-05
Update Dockerfile

Rombur:
This PR updates deal.II so you will need to pull the latest commit from the `adamantine` branch. I would also advise you to compile AborX when you update deal.II. It's not required yet but it will be in my next PR.


stvdwtt (approved):
LGTM

(Will merge pending CI check)

PR #84, opened 2021-02-26
Fix bug when using MPI

Rombur:
This is a hot fix for MPI simulation. I plan to add a MPI+CUDA test and probably add a sanitizer test. I spent hours finding a bug in the previous PR that I fixed in 10 minutes using the sanitizer. To review the PR, use the option to hide white space differences. I have a changed an if condition and the indentation is different.


stvdwtt (approved):
Looks good to me

PR #87, opened 2021-03-05
Fix bugs when using more than one MPI rank

Rombur:
Transferring solution when adding material is a little bit tricky. First, I forgot to update the ghost value. Second to use `CellDataTransfer`, we need data on every active including non-locally owned. So we need to be very careful on which operation is done on locally owned cells and which operation is done on active cells.


stvdwtt (approved):
Looks good and works for the demo problem.

PR #91, opened 2021-03-12
New way of calculating material properties on the fly

stvdwtt:
This changes where and how the material properties are calculated during the time step. Now the material properties and the heat source source term are calculated at each quadrature point in the local_apply. There are only two properties that cannot be calculated on the fly: the material id and the powder fraction. These are now stored in ThermalOperator, which has a method to extract them from the MaterialProperty object at the start of a simulation.

For the demo problem this is about 4x faster that the old way (~10x for the time step calculation since half is spent adding material).

Known issues:
1. ThermalOperatorDevice is nonfunctional and contains a ton of dummy code (but it does compile).
2. ThermalOperator::sync_stateful_material_properties isn't implemented. This is needed to output the material state in the VTK files. This doesn't impact the temperature field. In the output the state can be inferred from the temperature (except for the powder fraction).


application/adamantine.hh
```diff
@@ -661,10 +661,20 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   thermal_physics->compute_inverse_mass_matrix();
   dealii::LA::distributed::Vector<double, MemorySpaceType> solution;
   thermal_physics->initialize_dof_vector(initial_temperature, solution);
+
+  // I'm not sure if there's a better way to do this.
+  // extract_stateful_material_properties expects a MemorySpace::Host vector,
+  // but solution could be either Host or CUDA.
```
Rombur, line 667: `thermal_physics` is templated on the `MemorySpaceType` so you can deal with that there. You move the implementation in a free function that is directly used on the host and that is called after moving the data on the host when you are on the GPU.

source/MaterialProperty.cc
```diff
@@ -414,6 +421,72 @@ double MaterialProperty<dim>::compute_property_from_polynomial(
...
+dealii::VectorizedArray<double>
+MaterialProperty<dim>::compute_material_property(
+    StateProperty state_property,
+    std::array<dealii::types::material_id,
+               dealii::VectorizedArray<double>::size()>
+        material_id,
+    std::array<dealii::VectorizedArray<double>,
+               static_cast<unsigned int>(MaterialState::SIZE)>
+        state_ratios,
+    dealii::VectorizedArray<double> temperature) const
```
Rombur, line 435: This interface will never work on a GPU. We can't use `dealii::VectorizedArray` or `std::array`

source/MaterialProperty.cc
```diff
@@ -414,6 +421,72 @@ double MaterialProperty<dim>::compute_property_from_polynomial(
...
+        dealii::types::material_id m_id = material_id[n];
+
+        value[n] += state_ratios[material_state][n] *
+                    compute_property_from_polynomial(
+                        _state_property_polynomials.at(
+                            m_id)[material_state][property_index],
+                        temperature[n]);
+      }
+    }
+  }
```
Rombur, line 475: This whole implementation does not work on GPU. The `if ... else ...` means that the thread will diverge inside a warp.

source/MaterialProperty.hh
```diff
@@ -104,6 +112,27 @@ public:
    */
   dealii::DoFHandler<dim> const &get_dof_handler() const;
 
+  /**
+   * Compute a material property a quadrature point for a mix of states
```
Rombur, line 116:
```suggestion
   * Compute a material property at a quadrature point for a mix of states.
```

source/ThermalOperator.cc
```diff
@@ -15,22 +15,26 @@
...
+
 namespace adamantine
 {
 
 template <int dim, int fe_degree, typename MemorySpaceType>
 ThermalOperator<dim, fe_degree, MemorySpaceType>::ThermalOperator(
     MPI_Comm const &communicator,
-    std::shared_ptr<MaterialProperty<dim>> material_properties)
+    std::shared_ptr<MaterialProperty<dim>> material_properties,
+    std::vector<std::shared_ptr<HeatSource<dim>>> heat_sources)
```
Rombur, line 27:
```suggestion
    std::vector<std::shared_ptr<HeatSource<dim>>> const & heat_sources)
```

source/ThermalOperator.cc
```diff
@@ -187,34 +200,73 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::local_apply(
     // Store in a local vector the local values of src
     fe_eval.read_dof_values(src);
     // Evaluate only the function gradients on the reference cell
-    fe_eval.evaluate(false, true);
+    // fe_eval.evaluate(false, true);
+    fe_eval.evaluate(
+        true,
+        true); // Need the temperature to calculate the material properties
```
Rombur, line 206: Since we are touching these lines we should use the new version of this function (see [here](https://dealii.org/developer/doxygen/deal.II/classFEEvaluation.html#a710c252e9dccc0c4ce07044cac3bdd9b))

source/ThermalOperator.cc
```diff
@@ -187,34 +200,73 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::local_apply(
...
+      }
+      quad_pt_source *= inv_rho_cp;
+
+      fe_eval.submit_value(quad_pt_source, q);
+    }
+
     // Sum over the quadrature points.
-    fe_eval.integrate(false, true);
+    fe_eval.integrate(true, true);
     fe_eval.distribute_local_to_global(dst);
```
Rombur, line 250: Same here there is new version of this function.

source/ThermalOperator.cc
```diff
@@ -187,34 +200,73 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::local_apply(
...
 
 template <int dim, int fe_degree, typename MemorySpaceType>
 void ThermalOperator<dim, fe_degree, MemorySpaceType>::
-    evaluate_material_properties(
+    extract_stateful_material_properties(
         dealii::LA::distributed::Vector<double, dealii::MemorySpace::Host> const
             &temperature)
 {
-  // Update the state of the materials
+  // Update the state of the materials (is this needed here?)
```
Rombur, line 260: I don't remember. We should try removing it to see :smiley:

source/ThermalPhysics.hh
```diff
@@ -189,12 +197,14 @@ private:
    * Shared pointer to the material properties associated to the domain.
    */
   std::shared_ptr<MaterialProperty<dim>> _material_properties;
+
+  boost::property_tree::ptree _material_database;
```
Rombur, line 201: Missing comment.

Rombur (commented):
I think this is going in the right direction but significant work will be needed to have this PR work on the GPU. Let's talk about this tomorrow.

source/ThermalOperator.cc
stvdwtt: 😃
It isn't needed so I'm removing it.

application/adamantine.hh
```diff
@@ -135,7 +135,7 @@ inline void initialize_timers(MPI_Comm const &communicator,
 
 template <int dim, int fe_degree, typename MemorySpaceType,
           typename QuadratureType>
-std::vector<std::unique_ptr<adamantine::HeatSource<dim>>> &initialize(
```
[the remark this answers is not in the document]
stvdwtt: As we discussed in our last meeting the reason was lost with a lab notebook data loss, but since this way works we'll stick with it.

application/adamantine.hh
stvdwtt: It turns out that this can be completely side-stepped. That input vector isn't actually needed and is a holdover from an older implementation.

stvdwtt:
@Rombur, pending the CI check this is ready for you to take a look at. 

One thing I want to point out is that I'm only checking that the stateful material properties are correctly transferred from ThermalOperator to MaterialProperty for the case when they change for the entire domain in ThermalOperator. Originally I had hoped to select a few cells to change selectively, but I got tripped up on how to convert from the cell numbering in MaterialProperty to ThermalOperator and it seemed that I'd have to reimplement some of the code I'm checking in the test code. But if you think that type is necessary, I can add it. Perhaps unsatisfyingly, the output passes the eyeball test of where solid should turn to liquid in the DED demo.

Rombur:
Superseded by #109

PR #92, opened 2021-03-26
Add Caliper support to the build system

Rombur:
For now, we are measuring a very limited number of functions. We need to increase this overtime. Also let me know if you have a better method that having `#ifdef` through the code.


stvdwtt (approved):

stvdwtt (commented):
It looks like there's a build error for the CUDA build. I'm having a hard time parsing the logs to say something more specific.

Is `-pedantic` turning all of the "warning: style of line directive is a GCC extension" warnings into errors?

Rombur:
The error is a typo in the code:
```
/var/jenkins/workspace/adamantine/application/adamantine.cu(66): error: identifier "profiling_optional_database" is undefined
```
To see the log, go to artifact on the top right and then open pipeline.log. The warning is due to NVCC, it generates code that `pedantic` does not like.

PR #95, opened 2021-04-06
improve scan path robustness

stvdwtt:
When the simulation reaches the end of a scan path file, the heat source power is now set to zero and its position is the last position from the scan path file. Previously there could be unexpected behavior.

### DIFF
--- a/source/ScanPath.cc
+++ b/source/ScanPath.cc
@@ -164,6 +164,16 @@ void ScanPath::update_current_segment_info(
 
 dealii::Point<3> ScanPath::value(double const &time) const
 {
+  // If the current time is after the scan path data is over, return a point
+  // that is (presumably) out of the domain.
+  if (time > _segment_list.back().end_time)
+  {
+    dealii::Point<3> out_of_domain_point(std::numeric_limits<double>::max(),
+                                         std::numeric_limits<double>::max(),
+                                         std::numeric_limits<double>::max());
+    return out_of_domain_point;
+  }
+
   // Get to the correct segment (assumes that the current time is never
   // before the current segment starts)
   dealii::Point<3> segment_start_point;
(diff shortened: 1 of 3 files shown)


source/ScanPath.cc
```diff
@@ -177,7 +177,14 @@ dealii::Point<3> ScanPath::value(double const &time) const
...
           (time - segment_start_time);
 
-  return position;
+  if (time <= _segment_list.back().end_time)
+  {
+    return position;
+  }
+  else
+  {
+    return _segment_list.back().end_point;
```
Rombur, line 186: I think it makes sense to have something like
```c++
if (time > _segment_list.back().end_time)
  return _segment_list.back().end_point; 
```
has soon as we get in the function. What's the point on computing `position` since we don't use it. 

I think it may make sense to return an "invalid" point maybe something like `{-1e9, -1e9, -1e9}` instead of the last point. I am thinking about the case where we refine where the heat source will be. I am not sure that we deal correctly with a zero source, we may still refine something some cells when we shouldn't.

source/ScanPath.cc
```diff
@@ -188,7 +195,14 @@ double ScanPath::get_power_modifier(double const &time) const
...
   update_current_segment_info(time, segment_start_point, segment_start_time);
 
-  return _segment_list[_current_segment].power_modifier;
+  if (time <= _segment_list.back().end_time)
+  {
+    return _segment_list[_current_segment].power_modifier;
+  }
+  else
+  {
+    return 0.0;
```
Rombur, line 204: Same comment as above. Why bother computing `update_current_segment_info` in this case?

stvdwtt:
@Rombur, good points -- I've made updates to address those issues.

Rombur (approved):

PR #96, opened 2021-04-07
Calculate the "height" based on the scan path

stvdwtt:
Previously the "height" was calculated based on the top-most part of the mesh. This can cause some issues during material deposition (particularly if the material is placed on a new level while the heat source is still on the old level). 

The new implementation bases the "height" on the scan path instead of the mesh.


source/ThermalPhysics.templates.hh
```diff
@@ -332,6 +332,14 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::ThermalPhysics(
     else
       cell->set_active_fe_index(1);
   }
+
+  // Set the initial height of the heat source
+  double temp_height = -std::numeric_limits<double>::max();
```
Rombur, line 337:
```suggestion
  double temp_height = std::numeric_limits<double>::lowest();
```

source/ThermalPhysics.templates.hh
```diff
@@ -479,6 +472,15 @@ double ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
         dealii::LA::distributed::Vector<double, MemorySpaceType> &solution,
         std::vector<Timer> &timers)
 {
+
+  // Update the height of the heat source
+  double temp_height = -std::numeric_limits<double>::max();
```
Rombur, line 477:
```suggestion
  double temp_height = std::numeric_limits<double>::lowest();
```

Rombur (commented):
Thanks this is definitely an improvement compared to what we currently have. Two things to notice though:
 1. this doesn't work if different sources have different heights. I think that's not a problem right now but this something we need to keep in my mind.
 2. I think there is a bug if you have two sources and one of them just stops. In that case in #95 we return a source point that is out of the domain at a coordinate of `(max, max, max)`. Thus, the height will also be `max`. I think that the coordinate should be `(lowest, lowest, lowest)`.

stvdwtt:
Thanks for catching those -- I created #97 for item 1) and the new commit should fix item 2)

Rombur (approved):

PR #100, opened 2021-04-30
Material deposition that follows the scan path

stvdwtt:
This adds the functionality to automatically add material along the scan path. I've included the input files for a demonstration case in /test/data (demo_316_short.info and demo_316_short_scan_path.txt). @Rombur, I think that should be good for the Caliper test you mentioned.

Note 1: Currently this only works for a single heat source. I have a function to merge the material boxes from multiple heat sources, but right now it just picks the first one. I will add multi-heat-source support soon.

Note 2: This assumes that the scan path segments are always along one of the coordinate directions. We can relax that requirement, but with the resolutions that are currently practical we'd end up with pretty ugly zig-zags.


source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+      return read_material_deposition<dim>(geometry_database);
+    }
+    else
+    {
+      std::vector<
+          std::pair<std::vector<dealii::BoundingBox<dim>>, std::vector<double>>>
+          box_and_time_list;
+      for (auto const &source : heat_sources)
+      {
+        std::pair<std::vector<dealii::BoundingBox<dim>>, std::vector<double>>
```
Rombur, line 55:
```suggestion        
        auto
```

README.md
```diff
@@ -61,7 +61,14 @@ The following options are available:
...
   false (default value: false)
   * if material\_deposition is true:
-    * material\_deposition\_file: material deposition filename
+    * material\_deposition\_method: file, scan_paths
+    * if material\_deposition\_method is file:
+        * material\_deposition\_file: material deposition filename
+    * if material\_deposition\_method is scan_paths:
+        * deposition\_chunk\_length: length of material deposition boxes along the scan direction
+        * deposition\_width: width of material deposition boxes (in the plane of the material, normal to the scan direction, 3D only)
+        * deposition\_height: height of material deposition boxes (out of the plane of the material)
```
Rombur, line 70: Either use `deposition_chunk_*` or `deposition_*`. Don't mix both

source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
   {
     return {{}, {}};
   }
+  else
+  {
+    std::string method =
```
Rombur, line 36: Please add a `PropertyTreeInput` comment. If we remove this entry in the future, the serves as a reminder to update the README.

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+  // NOTE: Currently this assumes that scan path segments are aligned with the
+  // coordinate system.
+
+  std::pair<std::vector<dealii::BoundingBox<dim>>, std::vector<double>>
+      boxes_and_times;
+
+  // Load the box size information and lead time
+  double deposition_chunk_length =
+      geometry_database.get<double>("deposition_chunk_length");
+  double deposition_height = geometry_database.get<double>("deposition_height");
```
Rombur, line 151: See comment from the README.

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+{
+  // NOTE: Currently this assumes that scan path segments are aligned with the
+  // coordinate system.
+
+  std::pair<std::vector<dealii::BoundingBox<dim>>, std::vector<double>>
+      boxes_and_times;
+
+  // Load the box size information and lead time
+  double deposition_chunk_length =
+      geometry_database.get<double>("deposition_chunk_length");
```
Rombur, line 150: Here and below add `ProperTreeInput` comment.

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+  double deposition_height = geometry_database.get<double>("deposition_height");
+  double deposition_width =
+      geometry_database.get<double>("deposition_width", 0.0);
+
+  double lead_time = geometry_database.get<double>("deposition_lead_time");
+
+  // Loop through the scan path segements, adding boxes inside each one
+  std::vector<ScanPathSegment> segment_list = scan_path.get_segment_list();
+  double segment_start_time = 0.0;
```
Rombur, line 159: so what happens if the source is off and I turn it on after 1 second?

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+    if (segment.power_modifier > eps)
+    {
+      dealii::Point<3> segment_end_point = segment.end_point;
+      double segment_length = segment_end_point.distance(segment_start_point);
+      bool in_segment = true;
+      dealii::Point<3> center = segment_start_point;
+      double segment_velocity =
+          segment_length / (segment.end_time - segment_start_time);
+
+      // Set the segment orienation
```
Rombur, line 175:
```suggestion
      // Set the segment orientation
```

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+          else
+          {
+            center(1) = center(1) + center_increment;
+          }
+        }
+      }
+    }
+    segment_start_point = segment.end_point;
+    segment_start_time = segment.end_time;
+    ;
```
Rombur, line 275: Extra `;`

source/material_deposition.cc
```diff
@@ -97,6 +134,165 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+          }
+          else
+          {
+            center(1) = center(1) + center_increment;
+          }
+        }
+      }
+    }
+    segment_start_point = segment.end_point;
+    segment_start_time = segment.end_time;
```
Rombur, line 274: Does this work if we turn the source off and then back on? Can we do spot deposition?

Rombur (commented):
I have a few minor comments. One question does what does the 316 stands for in `demo_316`?

source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
   {
     return {{}, {}};
   }
+  else
```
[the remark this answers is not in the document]
stvdwtt: Fixed

source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+  {
+    std::string method =
+        geometry_database.get<std::string>("material_deposition_method");
+
+    ASSERT_THROW(method.compare("file") == 0 ||
```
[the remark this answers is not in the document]
stvdwtt: Fixed

source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+                     method.compare("scan_paths") == 0,
```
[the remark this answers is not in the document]
stvdwtt: Fixed

source/material_deposition.cc
```diff
@@ -29,7 +31,42 @@ read_material_deposition(boost::property_tree::ptree const &geometry_database)
...
+                 "Error: Method type for material deposition, '" + method +
+                     "', is not recognized. Valid options are: 'file' "
+                     "and 'scan_paths'");
+    if (method.compare("file") == 0)
```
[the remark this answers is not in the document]
stvdwtt: Fixed

source/material_deposition.cc
stvdwtt: Fixed

source/material_deposition.cc
stvdwtt: Fixed

source/material_deposition.cc
stvdwtt: Fixed

README.md
stvdwtt: My thought was that it would be helpful to make a distinction between the two. The `deposition_width` and the `deposition_height` directly control the width and heights of the track of material that is deposited. On the other hand, what I'm calling the `deposition_chunk_length` doesn't directly control the length of the track, that's controlled by the scan path. It is just the size that the track is quantized into.

At least that's what I was thinking when I named it. If you think making that distinction muddies things more than it clarifies, I can drop "chunk".

source/material_deposition.cc
stvdwtt: It should be fine (and I just tested to double check). Here I'm just setting the initial time for start of the scan path. From the format Alex gave me the scan path always starts right away and if you don't want to source to be on, you set the power to zero. More generally any time the source is off, the scan path still considers that to be a "segment", just one with zero power.

Does that make sense?

source/material_deposition.cc
stvdwtt: I also tested this one and we are able to do spot deposition. As with the above comment, the segments are continuous in time and the source being "off" corresponds to a zero power segment.

Or am I missing a failure mode?

stvdwtt:
Thanks for the comments and suggested edits. Perhaps opaquely, the "316" is from the type of steel that James is using for the builds (316 stainless steel) and is what the material parameters in the file correspond to.

README.md
Rombur: OK I understand. I would still drop the chunk. The truth is that you control the bounding box not the real track. This matters in the case where the track and the mesh are matching perfectly. In that case you want to have the bounding boxes smaller than the cells/the real track. If the track matches perfectly the cells, you will add material in the neighboring cells because the faces are activated and so the entire cell gets activated. That's without numerical error which makes the problem even worse.

stvdwtt: That makes sense, I'll drop "chunk".

Rombur (approved):

PR #104, opened 2021-05-24
WIP: Internal Dirichlet BCs

stvdwtt:
This PR adds the capability to ThermalPhysics to set a Dirichlet BC along the active/nonactive element boundary based on an input temperature field.

@Rombur, would you mind taking a quick look at this to see if this is along the lines of what you were expecting? I'm not happy with the test -- it just checks that the L1 norm decreases when the a Dirichlet BC of a lower value than the initial solution is applied. I'm also looking into a potential bug visible in the output file from the test case I added. I'll keep working on it, but I wanted to make sure that my general approach was sound.

### DIFF
--- a/source/ThermalPhysics.templates.hh
+++ b/source/ThermalPhysics.templates.hh
@@ -740,6 +740,45 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
 
   return solution;
 }
+
+template <int dim, int fe_degree, typename MemorySpaceType,
+          typename QuadratureType>
+void ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
+    set_internal_dirichlet_bcs(
+        dealii::LinearAlgebra::distributed::Vector<double> &solution,
+        dealii::LinearAlgebra::distributed::Vector<double> &imposed_temperature)
+{
+  // Loop over the faces looking for the interface between the two FE
+  // types
+  for (auto const &cell : dealii::filter_iterators(
+           _dof_handler.active_cell_iterators(),
+           dealii::IteratorFilters::LocallyOwnedCell(),
+           dealii::IteratorFilters::ActiveFEIndexEqualTo(0)))
+  {
+    for (const unsigned int face_no : cell->face_indices())
+    {
+      if (cell->face(face_no)->n_active_fe_indices() > 1)
+      {
+        std::vector<dealii::types::global_dof_index> indices;
+        auto dofs_per_face =
+            _dof_handler.get_fe_collection().max_dofs_per_face();
+        indices.resize(dofs_per_face);
+        cell->face(face_no)->get_dof_indices(indices, 0);
+
+        // Loop over the DoF on the face
+        for (auto dof : indices)
+        {
+          _affine_constraints.add_line(dof);
+          _affine_constraints.set_inhomogeneity(dof, imposed_temperature[dof]);
+          solution[dof] = imposed_temperature[dof];
+        }
+      }
+    }
+  }
+
+  _thermal_operator->reinit(_dof_handler, _affine_constraints, _q_collection);
+}
+
 } // namespace adamantine
 
 #endif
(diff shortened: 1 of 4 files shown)


source/ThermalPhysics.templates.hh
```diff
@@ -740,6 +740,45 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
 
   return solution;
 }
+
+template <int dim, int fe_degree, typename MemorySpaceType,
+          typename QuadratureType>
+void ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
+    set_internal_dirichlet_bcs(
```
Rombur, line 747: This function does the job for the internal BC but what about the final layer? Are you going to have a different function for that case? I feel like you could treat both cases here. Otherwise, it looks good.

stvdwtt: Thanks @Rombur, I hadn't thought about the final layer -- I'll roll that into this.

stvdwtt:
Hi @Rombur, I'm still getting some odd results from the test case that I wrote -- do you mind taking another look?

When I create the solution vector and then call `set_internal_dirichlet_bcs`, the vector is correct (both via the test and by manually visualizing the output) with the imposed value along the interface and the "original" value elsewhere. However, the solution after I take a couple time steps is not correct. The interface value remains the imposed value but the values at the vertices directly below those are higher than either the imposed value or the original value.

Two test cases:

Test 1:
Original value: 1000.
Imposed value: 1.

Test 2:
Original value: 1000.
Imposed value: 1000.

In both cases the imposed temperature along the interface is correct, but the temperature in the bulk material is the same wrong values for both tests (1021.25, then 999.375, then 1000.0 at the bottom boundary). In particular it is strange that I'm seeing any change in the solution for Test 2.

I moved the liquidus and solidus to be much higher temperatures so that there aren't any phase change effects. The numbers are different if I decrease the time step, but the general trend is the same. I also made sure that I'm well below the maximum stable time step.

Am I incorrectly applying the constraints in the time stepping loop?

source/ThermalPhysics.templates.hh
```diff
@@ -740,6 +740,52 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
...
+        dealii::LinearAlgebra::distributed::Vector<double> &solution,
+        dealii::LinearAlgebra::distributed::Vector<double> &imposed_temperature)
+{
+  // Loop over the faces looking for the interface between the two FE
+  // types or the top of the domain
+  _affine_constraints.clear();
```
Rombur, line 753: You clear the `AffineConstraints` but what if there are hanging nodes.

Rombur:
Imposing Dirichlet BC with matrix-free is tricky did you take a look at https://www.dealii.org/developer/doxygen/deal.II/step_37.html#InhomogeneousDirichletboundaryconditions

source/ThermalPhysics.templates.hh
stvdwtt: I wasn't sure the best way to treat that case -- if I didn't clear them first I was getting an exception about the constraints not being sorted. For better or worse, I was planning to address that once I got it working with a regular grid.

Since we will have to turn the Dirichlet BCs on and off for the data assimilation, does it make sense to have a separate `AffineConstraints` object for the Dirichlet BCs that we merge with an `AffineConstraints` object for the hanging nodes?

Rombur: Yeah if you take a look at the link I wrote, they are using two different `AffineConstraints` to impose the BC. I think that's the way to go.

stvdwtt: Sounds good -- I'm reading through that now.

Rombur:
@stvdwtt do you want to keep this open? Since we have established that we don't need this, I think we can close this.

stvdwtt:
@Rombur, yes, I think we can close it.

PR #105, opened 2021-05-27
Fix bug when adding material based on the scan path

stvdwtt:
There were a few sign errors in the code to determine the material to add from the scan path. I also split the add material timers into to (one to decide which material to add, one to actual activate it).


Rombur (approved):
Good catch

PR #106, opened 2021-05-28
AMR + Material Deposition

stvdwtt:
A few different fixes and added options for mixing AMR with adding material

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -516,7 +516,6 @@ void refine_mesh(
 
   for (unsigned int i = 0; i < n_beam_refinements; ++i)
   {
-
     // Compute the cells to be refined.
     std::vector<typename dealii::parallel::distributed::Triangulation<
         dim>::active_cell_iterator>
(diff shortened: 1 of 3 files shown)


application/adamantine.hh
```diff
@@ -511,18 +516,39 @@ void refine_mesh(
...
+      if (coarsen_after_beam || cell->level() < max_level)
         cell->set_refine_flag();
+    }
+
+    // Flag the cells for coarsening.
+    if (coarsen_after_beam)
+    {
+      for (auto cell : dealii::filter_iterators(
+               triangulation.active_cell_iterators(),
+               dealii::IteratorFilters::LocallyOwnedCell()))
```
Rombur, line 543: This is code somewhat complicated for no reason. What about this, you set the flag `set_coarsen_flag()` everywhere and then you loop over the cells to refine like this:
```C++
for (auto &cell : cells_to_refine)
{
  if (coarsen_after_beam)
    cell->clear_refine_flag();
  if (cell->level() < max_level)
    cell->set_refine_flag();
}
```

Rombur (changes requested):
Have you checked that this work? I have never tried to coarsen a cell which is already at the coarsest level.

application/adamantine.hh
stvdwtt: Good point -- I updated it along those lines.

(I also added an explicit check for if the level was greater than 0 before adding the coarsening flag. It did work without it -- there must be some internal check in Triangulation::execute_coarsening_and_refinement -- but it can't hurt to be more explicit and this allows us to add a minimum refinement later on if we wanted to.)

Rombur (approved):

PR #107, opened 2021-06-03
Fix radiative and convective BCs

stvdwtt:
Before this, `ThermalOperator::vmult_add` did not include the the reference temperature for convective or radiative boundary conditions. Therefore those boundary conditions drove the temperature of the system to 0 K.

This PR modifies `ThermalOperator::vmult_add` for non-zero reference temperatures (`conv_temperature_infty` and `rad_temperature_infty`).

Two new tests were added for `ThermalPhysics` to show that a block of material cooler than its environment will increase in temperature closer to the environmental temperature. 

The "spmv_rad" test for `ThermalOperator` was also modified. The reference solution assumes `rad_temperature_infty=0` (even though `rad_temperature_infty` is set to a non-zero value) and so it failed with the fix to `ThermalOperator::vmult_add`.  With some loss of generality, I set `rad_temperature_infty=0` for the test so that the old reference solution was still valid.

### DIFF
--- a/source/ThermalOperator.cc
+++ b/source/ThermalOperator.cc
@@ -210,12 +210,13 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::vmult_add(
             {
               for (unsigned int q = 0; q < n_face_q_points; ++q)
               {
-                cell_dst[i] -=
-                    inv_rho_cp *
-                    (conv_heat_transfer_coef + rad_heat_transfer_coef) *
-                    fe_face_values.shape_value(i, q) *
-                    fe_face_values.shape_value(j, q) * cell_src[j] *
-                    fe_face_values.JxW(q);
+                cell_dst[i] -= inv_rho_cp * fe_face_values.shape_value(i, q) *
+                               fe_face_values.shape_value(j, q) *
+                               (conv_heat_transfer_coef *
+                                    (cell_src[j] - conv_temperature_infty) +
+                                rad_heat_transfer_coef *
+                                    (cell_src[j] - rad_temperature_infty)) *
+                               fe_face_values.JxW(q);
               }
             }
           }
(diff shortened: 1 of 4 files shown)


source/ThermalOperator.cc
```diff
@@ -210,12 +210,13 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::vmult_add(
...
-                cell_dst[i] -=
-                    inv_rho_cp *
-                    (conv_heat_transfer_coef + rad_heat_transfer_coef) *
-                    fe_face_values.shape_value(i, q) *
-                    fe_face_values.shape_value(j, q) * cell_src[j] *
-                    fe_face_values.JxW(q);
+                cell_dst[i] -= inv_rho_cp * fe_face_values.shape_value(i, q) *
+                               fe_face_values.shape_value(j, q) *
+                               (conv_heat_transfer_coef *
+                                    (cell_src[j] - conv_temperature_infty) +
```
Rombur, line 216: I don't think that's correct. I have moved this term to the rhs. You can see it's [here](https://github.com/adamantine-sim/adamantine/blob/2862ed942234f4c93e10a7708bc82cae7feb53b9/source/ThermalPhysics.templates.hh#L691-L695). I guess the question is why that term is not working correctly.

stvdwtt: Ok, I can look into that to see if I can determine what's going on in the RHS.

If I/we can't figure that out before Tuesday, do you think the method of including it in the LHS is ok for forward Euler? It yields the correct qualitative behavior.

Rombur: Yes that would be fine.

Rombur:
@stvdwtt did you find out what's happening here?

stvdwtt:
I've been focusing on the data assimilation test case, but I can take a look at this next week.

Rombur:
I think I understand part of the problem. This is due to different expectations of what `vmult_add` does. Because part of the operator is moved to the source, it is missing from `vmult_add` and so calling `vmult_add` alone does not give the expected result. To get the entire operator, you need to call `evaluate_thermal_physics`. This explains the problem in the test.

stvdwtt:
@Rombur, that makes sense for the tests that I wrote in this PR. I think the investigation of what's going on with the boundary conditions essentially needs to start over, since I fundamentally misunderstood how they were implemented for the modifications and tests here.

Rombur:
Superseded by #109

PR #108, opened 2021-08-03
Add array for CPU and GPU

Rombur:
This PR adds new data structures: `Array2D`, `Array3D,` `Array4D`, and `Array5D` The memory space where the data is stored is determined by the last template parameters `MemorySpaceType`. The dimensions of the `Array` can be set at compile time using a template parameter or at run-time. In that last case, the template parameter should be set to `-1`. Note that I didn't implement all the combinations of compile/run-time parameters. I've implemented the one that I use in other parts of the code.

### DIFF
--- a/source/ArrayMD.hh
+++ b/source/ArrayMD.hh
@@ -1,0 +1,1224 @@
+/* Copyright (c) 2021, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
+
+#ifndef MDARRAY_HH
+#define MDARRAY_HH
+
+#include <utils.hh>
+
+// This file contains multiple arrays of different dimensions. The size of the
+// arrays can be set at compile-time or at runtime, similar to a Kokkos::View.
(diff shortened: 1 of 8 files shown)


source/ArrayMD.hh
```diff
@@ -0,0 +1,1224 @@
...
+                                                           unsigned int j,
+                                                           unsigned int k,
+                                                           unsigned int l) const
+{
+  return _values[i * (dim_1 * dim_2 * _dim_3) + j * (dim_2 * _dim_3) +
+                 k * _dim_3 + l];
+}
+
+/**
+ * Five-dimensional array with akk dimension set at compile-time.
```
stvdwtt, line 903:
```suggestion
 * Five-dimensional array with all dimensions set at compile-time.
```

stvdwtt (commented):
Looks great to me other than a minor typo in a comment.

stvdwtt:
@Rombur, was that comment the only thing you changed in the push an hour ago? I can't figure out how to see a diff from the original commit from yesterday.

Rombur:
Yes, sorry I just amended the last commit since it was a small change.

stvdwtt:
Sounds good -- just wanted to double-check.

PR #114, opened 2021-09-09
Ray tracing

Rombur:
This PR has three commits:
 1. Replace `local_size` which has been deprecated in deal.II with `locally_owned_size`
 2. Rename and rework the functions to read and set experimental data. In particular, I've fixed the way the data is set, instead of looking at each simulation point and searching the closest data point, I do the opposite. This way only the simulation points that are the closest to a data point gets set. The other entries are left untouched. I also return the result from ArborX so that you can use it to build H.
 3. Add the new ray tracing capabilities.


source/experimental_data.cc
```diff
@@ -158,39 +228,283 @@ void set_with_experimental_data(
...
-  dealii::ArborXWrappers::BVH bvh(points_values.points);
-  dealii::ArborXWrappers::PointNearestPredicate pt_nearest(support_points, 1);
+  dealii::ArborXWrappers::BVH bvh(support_points);
+  dealii::ArborXWrappers::PointNearestPredicate pt_nearest(points_values.points,
+                                                           1);
   auto [indices, offset] = bvh.query(pt_nearest);
 
-  // Fill in temperature
-  unsigned int const n_queries = support_points.size();
+  // Fill in the temperature
```
stvdwtt, line 236: Why do we need to fill in this temperature vector? Perhaps I'm misunderstanding it, but it seems like this temperature vector is for the whole simulated mesh and the code below is replacing each of the entries with the closest experimental point. I thought the plan was to create a vector with a the length of the number of observation points and then a map/pair relating the index from the vector to the DoF index on the simulation mesh.

source/experimental_data.cc
```diff
@@ -158,39 +228,283 @@ void set_with_experimental_data(
...
+        // Compute the distance along the ray direction between the origin of
+        // the ray and the intersection point.
+        auto const cross_product = dealii::cross_product_3d(edge_01, edge_02);
+        auto const &ray_origin = _rays_all_frames[frame][i].origin;
+        dealii::Tensor<1, dim> p0_ray({ray_origin[0] - point_0[0],
+                                       ray_origin[1] - point_0[1],
+                                       ray_origin[2] - point_0[2]});
+        double d = cross_product * p0_ray / det;
+        // We can finally compute the intersection point. It is possible that a
+        // ray intersects multiple face. For instance if the mesh is a cube the
```
stvdwtt, line 430:
```suggestion
        // ray intersects multiple faces. For instance if the mesh is a cube the
```

source/experimental_data.cc
```diff
@@ -9,21 +9,90 @@
...
+   * Return the `i`th Ray stored in the object.
+   */
+  Ray<3> const &get(unsigned int i) const { return _rays[i]; }
+
+private:
+  std::vector<Ray<3>> _rays;
+};
+} // namespace adamantine
+
+namespace ArborX
```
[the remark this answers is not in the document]
Rombur: This is only used by ArborX in the `.cc`. There is no point to export that everywhere in the code.

source/experimental_data.cc
Rombur: No that's not what it does. That's what it used to do. Now, I only partially fill the `temperature` vector with the experimental data. This means that each observation point is associated to a single simulation point. The function returns the data to build the map. I can return the map directly if you prefer. 
I don't see the point of filling a vector with the length of the observation point since that's the input of the function (`points_values.values` is exactly that).

stvdwtt: Got it, thanks. I think between `{indices, offset}` and `points_values.values`, there's enough to build the map the observation temperatures to the simulation DoFs. I'm not seeing where I'd need the partially filled temperature vector, except maybe for visualization. Or am I missing something?

Rombur: Right. It's there because at first we thought that we would impose the experimental data using Dirichlet condition and I've let it because I think it is useful for debug purpose.

stvdwtt (approved):

PR #116, opened 2021-09-29
Add capability for (identical) ensemble simulations

stvdwtt:
This adds the capability for ensemble simulations. Currently they are all identical, but the ability to vary parameters will be added in a future PR.

Note: This has an odd issue where I can't call `output_pvtu' directly on an element of a vector of `PostProcessor`, instead I have to make a copy. PostProcessor doesn't need a persistent state, so this works.


Rombur:
It looks like there is an out-of-bound access triggered by `test_integration_2d` See [here](https://cloud.cees.ornl.gov/jenkins-ci/job/adamantine/111/testReport/junit/projectroot/tests/Test___clang_sanitizer___test_integration_2d_1/).

application/adamantine.hh
```diff
@@ -875,4 +875,285 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
+  std::vector<dealii::AffineConstraints<double>> affine_constraints_ensemble;
+  for (unsigned int member = 0; member < ensemble_size; ++member)
+  {
+    affine_constraints_ensemble.push_back(
+        thermal_physics_ensemble[member]->get_affine_constraints());
+
+    // Segfaults if I pass post_processor_ensemble[member] directly to
+    // output_pvtu, unclear why
+    auto post_processor_member = post_processor_ensemble.at(member);
+    output_pvtu(post_processor_member, cycle, n_time_step, time,
```
Rombur, line 969: The code segfaults on my machine here.

stvdwtt: Thanks, I must just be getting lucky on my machine that when I do the copy the right thing is in the right place in memory. I'll see how I can rework it, and hopefully reproduce the segfault on my machine.

Rombur: That's probably why the sanitizer complain. One possible issue is that everything is duplicated except `Geometry`. You only have a single geometry for all the ensembles. That should be fine but since the error seems to be related to the Triangulation it might be worth to duplicate the `Geometry` too.

Rombur:
Still failing :/ You know that there is clang on the docker image. You can take a look here how to get the sanitizer to work:
https://github.com/adamantine-sim/adamantine/blob/master/ci/jenkins_config#L50-L71

Just make sure that the path to llvm-symbolizer is correct.

stvdwtt:
@Rombur, thanks -- with that I was able to reproduce the segfault locally. Changing the postprocessor vector to a vector of unique_ptrs seems to work locally, although I'm not sure why it's different.

application/adamantine.hh
```diff
@@ -875,4 +875,290 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
+  std::vector<std::unique_ptr<adamantine::Geometry<dim>>> geometry_ensemble;
+
+  std::vector<std::unique_ptr<adamantine::PostProcessor<dim>>>
+      post_processor_ensemble;
+
+  for (unsigned int member = 0; member < ensemble_size; ++member)
+  {
+
+    geometry_ensemble.push_back(std::unique_ptr<adamantine::Geometry<dim>>(
+        new adamantine::Geometry<dim>(communicator, geometry_database)));
```
Rombur, line 941: Instead of 
```C++
geometry_ensemble.push_back(std::unique_ptr<adamantine::Geometry<dim>>(
        new adamantine::Geometry<dim>(communicator, geometry_database)));
```
You can do 
```C++
 geometry_ensemble.push_back(std::make_unique<adamantine::Geometry<dim>>(
        (communicator, geometry_database)));
```
It does the same thing but you have less to type.

Rombur (approved):
This looks good but I have a question about the overall design. Right now, you could have save a lot of code doing something like:
```c++
void run_ensemble()
{
  for (int =0; i<n_ensembles; ++i)
    run(..., i);
}
```
You would just have to modify `run` to take the extra arguments. So where do you plan to add the data assimilation part?

application/adamantine.hh
stvdwtt: My plan was to add the data assimilation inside the time stepping loop. The EnKF needs all of the ensemble members to be at the same point in time when assimilating and then after assimilation all of the solution field change.

I could imagine maybe doing something like you wrote between data assimilation points, but then we'd have to track the state in some persistent way outside of `run`/`run_ensemble`.

PR #121, opened 2021-10-19
Only use parallel std library when using GLIBCXX

Rombur:
`libc++` does not support parallel std library. This guards the use of parallel std library, so that we only use it with `libstdc++`. The other solution is to just not use the parallel std. I don't have a strong opinion either way. I think it is nice to use it by the macros make the code ugly and we probably don't gain much by using this.

### DIFF
--- a/source/DataAssimilator.cc
+++ b/source/DataAssimilator.cc
@@ -10,7 +10,10 @@
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/lac/linear_operator_tools.h>
 
+// libc++ does not support parallel std library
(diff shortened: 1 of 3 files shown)


Rombur:
It looks nicer but it doesn't work because you would need to include the comma otherwise it will expand to `std::transform( , xxx)`. However, this now looks very strange because the argument following the macro is not separated by a comma. The nicest solution is to create a wrapper for the `std` functions we are using but that's a little bit overkilled. Let's merge this and we can revisit later. I don't think it's worth spending too much time on this.

PR #124, opened 2021-10-21
Add data assimilation capabilities to the app

stvdwtt:
This PR adds data assimilation capabilities to the adamantine app.

Main changes:
- Adds the option for the initial temperature to be different for different ensemble members
- Adds a section of code before the time stepping loop to read experimental data (if directed by the input file)
- Fix how the experiment data size is set in DataAssimilator
- Add data assimilation to the main time stepping loop
- Add an integration test for data assimilation

Other changes:
- Fix parsing of the experimental log file (previously the first line was discarded)
- Add section break comments in `run_ensemble` to make it easier to read through (at least for me)
- Add some output to screen about the progress of experimental data parsing and data assimilation

Notes:
- Data assimilation still only works in serial
- Data assimilation still uses some dense linear algebra


Rombur:
There is a bug with `rank`, you need to get it before using it (`dealii::Utilities::MPI::this_mpi_process(comm)` will give to you). It also look like clang-sanitizer is not happy.

PR #128, opened 2021-11-05
Add direction of material deposition

Rombur:
This is the first PR to support deposition of anisotropic material. This PR adds support for reading the deposition angle when we add material using the bounding boxes. When using the scan path to add material, I compute the angle, so there is no change in the input file. Two  more things:
1. in 2D, you need to provide the deposition angle but right now I don't use it. I don't know what it means to have a deposition angle in 2D. Should I get ride of it,  use it, let the code as is?
2. I wonder if there is a mistake in the README. We say that the unit of the scan path is in `mm` but the tests are done using values of the order of `10^-3`. This makes me think that the values are in meters.


stvdwtt (approved):
I think explicitly including the deposition angle in 2D is the right approach. I'm not sure if I'd use it this way, but one could think of two kinds of 2D simulations. One has the deposition head moving in-plane where we see the whole scan path. The second has the deposition head moving out of plane and we only see where it crosses the plane we're simulating.

PR #129, opened 2021-11-09
Small fixes

Rombur:
- Fix the READMS
- Fix scan path input files
- Read number of segments of the scan path (this easier to read if you use `Hide Widespace`). We know stop reading the file when the file ends or we reached the maximum number of segments to read, whichever comes first.
- Support `csv` format for event scan path

### DIFF
--- a/source/ScanPath.cc
+++ b/source/ScanPath.cc
@@ -123,7 +123,7 @@ void ScanPath::load_event_series_scan_path(std::string scan_path_file)
     ScanPathSegment segment;
 
     std::vector<std::string> split_line;
-    boost::split(split_line, line, boost::is_any_of(" "),
+    boost::split(split_line, line, boost::is_any_of(" ,,"),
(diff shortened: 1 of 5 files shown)


source/ScanPath.cc
```diff
@@ -119,7 +123,7 @@ void ScanPath::load_event_series_scan_path(std::string scan_path_file)
     ScanPathSegment segment;
 
     std::vector<std::string> split_line;
-    boost::split(split_line, line, boost::is_any_of(" "),
+    boost::split(split_line, line, boost::is_any_of(" ,,"),
```
stvdwtt, line 126: Is the double comma intentional?

Rombur: Yes, there are two separators `" "` and `,` and then must be separated by a comma. At first, I had `" ,\," ` to make things clearer but it gave me warning

stvdwtt (approved):

Issue #130, opened 2021-11-09
Segfault using multithreading


Rombur:
We have a segfault in the code in the code using multithreading. I don't know what triggers it but it seems to happen in `cell_local_apply`.  Some input files will trigger the segfault using a single MPI rank while others fail when using multiple MPI ranks (`./adamantine -i demo_316_short.info` passes but `mpirun -np 2 ./adamantine -i demo_316_short.info` fails). Here is a backtrace using clang-sanitizer address. Using `DEAL_II_NUM_THREADS=1` fixes the problem. Since we are using MPI for parallelization on the host, I propose that we always set `DEAL_II_NUM_THREADS=1` in the Docker image and update the `README`. Multithreading in deal.II is done using TBB but it will be removed in the future. Once TBB has been replaced we can revisit this.
```
AddressSanitizer:DEADLYSIGNAL                                                                                                                                                                                                                =================================================================                                                                                                 
==1773==ERROR: AddressSanitizer: SEGV on unknown address 0x6248000108d8 (pc 0x0000014ddeec bp 0x7fdd74eb2830 sp 0x7fdd74eb2750 T89)                                                                                                          ==1773==The signal is caused by a READ memory access.                               
AddressSanitizer:DEADLYSIGNAL                                                                                                                                    
AddressSanitizer:DEADLYSIGNAL                                                                                                                                                                                             
AddressSanitizer:DEADLYSIGNAL                                                                                                                                    
    #0 0x14ddeec in adamantine::ScanPath::update_current_segment_info(double, dealii::Point<3, double>&, double&) const /home/dev/adamantine/build/../source/ScanPath.cc:159:62
    #1 0x14deccf in adamantine::ScanPath::value(double const&) const /home/dev/adamantine/build/../source/ScanPath.cc:184:3                        
    #2 0x1cf7317 in adamantine::GoldakHeatSource<3>::value(dealii::Point<3, double> const&, double, double) const /home/dev/adamantine/build/../source/GoldakHeatSource.cc:34:59
    #3 0x164e7f0 in adamantine::ThermalOperator<3, 3, dealii::MemorySpace::Host>::cell_local_apply(dealii::MatrixFree<3, double, dealii::VectorizedArray<double, 2ul> > const&, dealii::LinearAlgebra::distributed::Vector<double, dealii::Me
[...]

PR #132, opened 2021-11-16
Disable multithreading in Docker image and update README

Rombur:
I was trying to add a new test case but I hit the multithreading bug https://github.com/adamantine-sim/adamantine/issues/130 This PR updates the Docker image and the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,13 @@ mpirun -n 2 ./adamantine --input-file=input.info
 Note that the name of the input file is totally arbitrary, `my_input_file` is as
 valid as `input.info`.
 
+There is a [known bug](https://github.com/adamantine-sim/adamantine/issues/130)
+when using multithreading. To deactivate multithreading use
+```bash
+export DEAL_II_NUM_THREADS=1
+```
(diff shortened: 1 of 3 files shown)


stvdwtt (approved):

Rombur:
Retest this please

Rombur:
The reason the failed is due to the fact that we were unlucky in the choice of the random generator and that we converge to the correct solution only for a large sampling size. We use random number in `ensemble_management` and `DataAssimilator`. In `DataAssimilator` the seed is always the same but in `ensemble_management` the seed was different for each run that means that it is impossible to reproduce the result between runs. I have changed that so that the seed is always the same.

stvdwtt:
Ok -- I'm a little leery of always using the same seed (in both cases really, if I was thinking about it I would have used a random seed in DataAssimilator too) for productions runs, but it should be ok for now.

Rombur:
Yes, the numbers will always be the same between run but they will still be random. It's quite common to fix the seed that way if you change a parameter you know that the difference in the output is due to that parameter.

PR #133, opened 2021-11-16
No full matrices in DataAssimilator for the random number fill for diagonal R

stvdwtt:
This removes one of the places where full matrices are used for data assimilation. The diagonal R case is the only case currently supported by the adamantine application, but we still have the non-diagonal case available as a more general fall-back.

### DIFF
--- a/source/DataAssimilator.cc
+++ b/source/DataAssimilator.cc
@@ -59,6 +59,14 @@ void DataAssimilator::update_ensemble(
   adamantine::ASSERT_THROW(_expt_size == expt_data.size(),
                            "Error: Unexpected experiment vector size.");
 
+  // Check if R is diagonal, needed for filling the noise vector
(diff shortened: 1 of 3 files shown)


source/DataAssimilator.cc
```diff
@@ -59,6 +59,14 @@ void DataAssimilator::update_ensemble(
   adamantine::ASSERT_THROW(_expt_size == expt_data.size(),
                            "Error: Unexpected experiment vector size.");
 
+  // Check if R is diagonal, needed for filling the noise vector
+  auto bandwidth = R.get_sparsity_pattern().bandwidth();
+  bool R_is_diagonal = false;
+  if (bandwidth == 0)
+  {
+    R_is_diagonal = true;
+  }
```
Rombur, line 68:
```suggestion
  bool const R_is_diagonal = bandwidth == 0 ? true : false;
```

Rombur (approved):
Looks good.

PR #135, opened 2021-11-19
Finish anisotropy

Rombur:
This is the last PR for the anisotropy. It propagates the cosine and the sine from the material through the rest of the code and it uses it in `ThermalOperator` and `ThermalOperatorDevice`


application/adamantine.hh
```diff
@@ -828,23 +851,15 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
                 << std::endl;
-    // Remove elements that have been activated
-    elements_to_activate.erase(elements_to_activate.begin() + activation_start,
-                               elements_to_activate.begin() + activation_end);
-    deposition_times.erase(deposition_times.begin() + activation_start,
-                           deposition_times.begin() + activation_end);
-
-    material_deposition_boxes.erase(
-        material_deposition_boxes.begin() + activation_start,
-        material_deposition_boxes.begin() + activation_end);
```
stvdwtt, line 847: Removing this block is unrelated to the other changes, right? It's just that adding the `if` statement on line 854 means we don't have to remove the elements/times/deposition boxes that we already used?

stvdwtt (commented):
Overall looks good, just a few minor comments/questions.

application/adamantine.hh
Rombur: That's correct. We could still remove the elements but there is little point in doing it. I prefer to remove the code, so it's one less place to have a bug.

source/ThermalPhysics.templates.hh
```diff
@@ -524,7 +524,13 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::ThermalPhysics(
     // If the center of the cell is below material_height, it contains material
     // otherwise it does not.
     if (cell->center()[axis<dim>::z] < material_height)
+    {
       cell->set_active_fe_index(0);
+      // Set material deposition cos and sin. We arbitrarily choose cos = 1 and
+      // sin = 1
+      _deposition_cos.push_back(1.);
+      _deposition_sin.push_back(0.);
+    }
```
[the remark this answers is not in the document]
Rombur: We have already so many parameters. I would prefer if we waited until we get a case where we need this to modify these values before add more parameters

stvdwtt (approved):

Rombur:
I don't know why the test is failing. It passes on when I run the container locally. The only thing that's different is the machine where the tests are run. The tests pass on fetnat but fail on jiminy

Rombur:
jiminy uses a different kind of gpu than fetnat. jiminy uses V100s and fetnat uses V100. Not sure if it's the reason there is a problem but that's the only difference I know of.

Rombur:
I found the bug and I am in the process of fixing it. There are two out-of-bound accesses in some `std::vector` The faulty code is on the CPU so I don't understand why the sanitizer doesn't report an error.

Rombur:
This is ready to be merged but I need another review because I had to change quite a bit of code. Basically the problem was that the deposition cosine/sine was updated at the same than the matrix free. This works during initialization but when we add material or when the mesh is refined we first need to reinitialize the matrix free and then we can get the cosine/sine from the new mesh.

stvdwtt (approved):

PR #136, opened 2021-11-23
Verify input file

stvdwtt:
This adds a check to validate the input file at the start of the simulation. Most of the input checks in the code have been moved here, with a few exceptions. In cases where there is a defined list of options, the core part of the code still has some `else` statements with thrown exception if input doesn't match the options. I think this helps the readability of the code and doesn't add much complexity. The other case is when the validation is intricately tied to the parsing (e.g. checking for the experimental camera files). At least at this point, I didn't want to have to repeat too much of the logic/parsing process in two places.

To test the input validation, I tried to balance coverage, length, and my time. It isn't exhaustive (e.g. I don't consider all possible combinations of valid inputs), but it is pretty comprehensive. If there are important checks that I'm missing, I'm happy to add them here or in another PR.

I also updated the README to list more directly which input are required and which are not.

I did not change the default behavior for any of the inputs in the PR, although at some point we may want to reconsider which inputs have default values and which do not.

### DIFF
--- a/application/adamantine.cc
+++ b/application/adamantine.cc
@@ -7,6 +7,8 @@
 
 #include "adamantine.hh"
 
+#include <validate_input_database.hh>
+
 #ifdef ADAMANTINE_WITH_ADIAK
 #include <adiak.hpp>
 #endif
@@ -73,6 +75,9 @@ int main(int argc, char *argv[])
                              "The file " + filename + " does not exist.");
     boost::property_tree::ptree database;
     boost::property_tree::info_parser::read_info(filename, database);
+    adamantine::validate_input_database(database);
+
+    // TODO: Catch the exception and print out the message
 
 #ifdef ADAMANTINE_WITH_CALIPER
     cali::ConfigManager caliper_manager;
@@ -111,7 +116,6 @@ int main(int argc, char *argv[])
         database.get_child("geometry");
     // PropertyTreeInput geometry.dim
     int const dim = geometry_database.get<int>("dim");
-    adamantine::ASSERT_THROW((dim == 2) || (dim == 3), "dim should be 2 or 3");
 
     unsigned int rank = dealii::Utilities::MPI::this_mpi_process(communicator);
     if (rank == 0)
--- a/application/adamantine.cc
+++ b/application/adamantine.cc
@@ -75,9 +75,15 @@ int main(int argc, char *argv[])
                              "The file " + filename + " does not exist.");
     boost::property_tree::ptree database;
     boost::property_tree::info_parser::read_info(filename, database);
-    adamantine::validate_input_database(database);
-
-    // TODO: Catch the exception and print out the message
+    try
+    {
+      adamantine::validate_input_database(database);
+    }
+    catch (std::runtime_error const &exception)
+    {
+      std::cerr << exception.what();
+      return 1;
+    }
 
 #ifdef ADAMANTINE_WITH_CALIPER
     cali::ConfigManager caliper_manager;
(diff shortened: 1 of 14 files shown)


Rombur (approved):
Looks good can you just try the code I suggested

tests/test_validate_input_database.cc
```diff
@@ -0,0 +1,494 @@
...
+
+  filename = "bare_plate_L_ensemble.info";
+  boost::property_tree::info_parser::read_info(filename, database);
+  validate_input_database(database);
+
+  filename = "demo_316_short.info";
+  boost::property_tree::info_parser::read_info(filename, database);
+  validate_input_database(database);
+
+  filename = "integration_2D.info";
```
Rombur, line 67: This should be:
```suggestion
  filename = "integration_2d.info";
```
I am not sure why it passes in release mode.

Rombur: My bad it's failing in release mode too

stvdwtt: That's really odd/concerning that it worked in both debug and release mode on my computer...

Regardless, that typo is now fixed.

application/adamantine.cc
```diff
@@ -73,6 +75,9 @@ int main(int argc, char *argv[])
                              "The file " + filename + " does not exist.");
     boost::property_tree::ptree database;
     boost::property_tree::info_parser::read_info(filename, database);
+    adamantine::validate_input_database(database);
+
+    // TODO: Catch the exception and print out the message
```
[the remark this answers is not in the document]
stvdwtt: Added. (Slightly modified to add a new line, which makes it easier to parse when MPI complains about the non-zero exit code.)

tests/test_validate_input_database.cc
Rombur: Maybe the file existed on your computer. If you are on mac the filesystem is the file system is not case sensitive. If the file was mapped from the docker to the mac, maybe something weird happened.

application/adamantine.cc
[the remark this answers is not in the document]
Rombur: You can return zero it it looks better. No body will check the return code of main.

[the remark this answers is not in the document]
stvdwtt: Good idea, I just pushed that change.

Rombur:
Error from cuda
```
/var/jenkins/workspace/adamantine/application/adamantine.cu(77): error: namespace "adamantine" has no member "validate_input_database"
```

Rombur:
One of the test times out. I think the problem is due to the new machines that we are using for the CI. We don't control what kind of CPU we get so it's possible we got something slower. Can you change line 74 in `ci/jenkins_config` to
```
 sh 'ctest --timeout 1800 --no-compress-output -R test_ -T Test'
```

Rombur:
This doesn't help. I think the only way is to step in with a debugger to see what's going on. Do you know how to do it?

stvdwtt:
Is it possible that it is still legitimately timing out? The test `test_material_deposition_2` is a two-MPI-task version of the test, right? It looks like that took over 1000s.

```
18/23 Test #19: test_material_deposition_1 .......***Timeout 1800.11 sec
      Start 20: test_material_deposition_2
19/23 Test #20: test_material_deposition_2 .......   Passed  1071.34 sec
```

Rombur:
I guess so. I wonder why the time of that test has increased so much. The other tests run in about the same amount of time as they did a week ago. You can try increasing the time again and see it that fixes it.

Rombur:
We will need to investigate why this test is suddenly so slow. Since this capability is badly needed, I will merge it now and we can fix the test later on.

PR #137, opened 2021-12-01
Add localization for DA

stvdwtt:
This removes the last dense matrix operations in `DataAssimilator` by adding support for localization. As expected, this permits much larger meshes (at least 8x, I haven't tried larger yet), depending on the localization cutoff length.


stvdwtt:
I'm looking into the integration test error and will push a fix when I have a chance.

source/DataAssimilator.cc
```diff
@@ -288,44 +357,80 @@ void DataAssimilator::fill_noise_vector(dealii::Vector<double> &vec,
...
-    std::vector<VectorType> vec_ensemble) const
+double DataAssimilator::gaspari_cohn_function(double const r) const
 {
-  unsigned int num_ensemble_members = vec_ensemble.size();
-  unsigned int vec_size = 0;
-  if (vec_ensemble.size() > 0)
+  if (r < 1.0)
+  {
+    return 1. - 5. / 3. * r * r + 5. / 8. * r * r * r + 0.5 * r * r * r * r -
+           0.25 * r * r * r * r * r;
```
Rombur, line 365:
```suggestion
    return 1. - 5. / 3. * std::pow(r, 2) + 5. / 8. * std::pow(r, 3) + 0.5 * std::pow(r, 4) -
           0.25 * std::pow(r, 5);
```

source/DataAssimilator.cc
```diff
@@ -288,44 +357,80 @@ void DataAssimilator::fill_noise_vector(dealii::Vector<double> &vec,
...
+  }
+  else if (r < 2)
   {
-    vec_size = vec_ensemble[0].size();
+    return 4. - 5. * r + 5. / 3. * r * r + 5. / 8. * r * r * r -
+           0.5 * r * r * r * r + 1. / 12. * r * r * r * r * r - 2. / (3. * r);
```
Rombur, line 370: Same here

source/DataAssimilator.hh
```diff
@@ -22,6 +22,14 @@
 
 namespace adamantine
 {
+
+enum class LocalizationCutoff
```
Rombur, line 26: Can you add a comment about this enum and a reference for the gaspari-cohn function

tests/test_data_assimilator.cc
```diff
@@ -50,9 +50,9 @@ class DataAssimilatorTester
     boost::property_tree::ptree database;
     database.put("import_mesh", false);
     database.put("length", 1);
-    database.put("length_divisions", 2);
+    database.put("length_divisions", 1);
```
Rombur, line 53: Why did you change this?

tests/test_data_assimilator.cc
```diff
@@ -591,6 +673,49 @@ class DataAssimilatorTester
...
+      }
+    }
+
+    // FIXME This can be a problem for even moderately sized meshes
+    dealii::FullMatrix<double> cov(vec_size);
+
+    anomaly.mTmult(cov, anomaly);
+
+    return cov;
+  };
```
Rombur, line 718:
```suggestion
  }
```

tests/test_data_assimilator.cc
stvdwtt: Before the number of dofs on the mesh was different than `sim_size`. Without localization, it was fine that the mock simulation vector was shorter than what it should be. With localization on, the distance calculation in the localized covariance matrix calculation gets messed up if they aren't consistent.

Rombur:
Still times out

stvdwtt:
Odd -- I'll take a closer look at it this afternoon.

Rombur:
Still times out but it goes further. We have two choices:
 - We split the 3D integration tests in two file so that the time limit applies to each test and not the sum of them. We should also change https://github.com/adamantine-sim/adamantine/blob/master/ci/jenkins_config#L74 to use `-j2` or `-j4` otherwise it will take too long to run the CI.
 - We exclude the 3D integration tests when using clang. Adding something like `-E integration_3d` to https://github.com/adamantine-sim/adamantine/blob/master/ci/jenkins_config#L74 should do the trick.
 
What do you think?

stvdwtt:
I was able to trim the DA integration test runtime by a factor of two (at least on my machine), so hopefully that won't be an issue anymore. I also added the "-j 4" flag to the ctest command as you suggested. If we think there could still be some issues I can split up the integration tests, but for now I was going for the lightest touch that would work.

PR #138, opened 2022-01-28
Augmented state ensemble

stvdwtt:
This PR adds functionality for an augmented state ensemble for data assimilation, that is to add unobservable model parameters to the state vector. This requires changing the solution vector to a block vector and also requires allowing some model parameters to be able to change in the middle of a calculation.

Unrelated to the core of these changes, a bug in time incrementing for ensemble simulations was fixed. (Previously, each ensemble member update incremented the time by the time step, meaning that the time was incremented by num_members*dt rather than dt.)

NOTE: The current interface for LA:distributed:BlockVector does not allow the memory space to be identified. Therefore the BlockVectors in this PR live exclusively on the host.


Rombur:
test_integration_3d times out when using clang-sanitizer

README.md
```diff
@@ -186,10 +186,13 @@ The following options are available:
...
   * new\_material\_temperature\_stddev: The standard deviation for the temperature of material added during the process (default value: 0.0)
   * beam\_0\_max\_power\_stddev: The standard deviation for the max power for beam 0 (if it exists) (default value: 0.0)
+  * beam\_0\_absorption\_efficiency\_stddev: The standard deviation for the absorption efficiency for beam 0 (if it exists) (default value: 0.0)
 * data\_assimilation: (optional)
   * assimilate\_data : Whether to perform data assimilation (default value: false)
   * localization\_cutoff\_function : The function used to decrease the sample covariance as the relevant points become farther away: gaspari\_cohn, step\_function, none (default: none)
   * localization\_cutoff\_distance : The distance at which sample covariance entries are set to zero (default: infinity)
+  * augment\_with\_beam\_0\_absorption : Whether to augment the state vector with the beam 0 absorption efficiency (default: false)
+  * augment\_with\_beam\_0\_max_power : Whether to augment the state vector with the beam 0 max power (default: false)
```
Rombur, line 195: Looks like we have a mix of `xxx :` and  `xxx:` but we use the second option more often. Can you change the few lines that use the first option.

README.md
```diff
@@ -186,10 +186,13 @@ The following options are available:
   * initial\_temperature\_stddev: The standard deviation for the initial temperature of the material (default value: 0.0)
   * new\_material\_temperature\_stddev: The standard deviation for the temperature of material added during the process (default value: 0.0)
   * beam\_0\_max\_power\_stddev: The standard deviation for the max power for beam 0 (if it exists) (default value: 0.0)
+  * beam\_0\_absorption\_efficiency\_stddev: The standard deviation for the absorption efficiency for beam 0 (if it exists) (default value: 0.0)
```
Rombur, line 189: Does this work only for `beam_0`? It's fine for now but we need to find a better way to write the inputs

application/adamantine.hh
```diff
@@ -1030,22 +1039,88 @@ run_ensemble(MPI_Comm const &communicator,
...
+  {
+    // PropertyTreeInput data_assimilation
+    data_assimilation_database = data_assimilation_optional_database.get();
+
+    // PropertyTreeInput data_assimilation.assimilate_data
+    if (data_assimilation_database.get("assimilate_data", false))
+    {
+      assimilate_data = true;
+
+      unsigned int index = 0;
```
Rombur, line 1065: Why do you need `index`?

application/adamantine.hh
```diff
@@ -1030,22 +1039,88 @@ run_ensemble(MPI_Comm const &communicator,
...
+            solution_augmented_ensemble[member].block(1)[index] =
+                beam_0_absorption[member];
+          }
+          else if (augmented_state_parameters.at(index) ==
+                   adamantine::AugmentedStateParameters::beam_0_max_power)
+          {
+            solution_augmented_ensemble[member].block(1)[index] =
+                beam_0_max_power[member];
+          }
+        }
```
Rombur, line 1122: This is fine for now but it needs a `FIXME` that we need to generalized this code later.

application/adamantine.hh
```diff
@@ -1030,22 +1039,88 @@ run_ensemble(MPI_Comm const &communicator,
...
+
+      // PropertyTreeInput sources.beam_0.absorption_efficiency
+      database_ensemble[member].put("sources.beam_0.absorption_efficiency",
+                                    beam_0_absorption[member]);
+
+      // Populate the parameter augmentation block of the augmented state
+      // ensemble
+      if (assimilate_data)
+      {
+        solution_augmented_ensemble[member].block(1).reinit(
```
Rombur, line 1104: I think it would be nice to have somewhere in the function something like (maybe you can find better name):
```c++
int constexpr base_state = 0;
int constexpr augmented_state = 1;
```
This way we don't need to remember what's block 0 and block 1

application/adamantine.hh
```diff
@@ -1290,16 +1356,19 @@ run_ensemble(MPI_Comm const &communicator,
...
         (activation_start == activation_end) ? false : true;
 #endif
     timers[adamantine::evol_time].start();
     for (unsigned int member = 0; member < ensemble_size; ++member)
     {
-      time = thermal_physics_ensemble[member]->evolve_one_time_step(
-          time, time_step, solution_ensemble[member], timers);
+      temp_time = thermal_physics_ensemble[member]->evolve_one_time_step(
+          time, time_step, database_ensemble[member].get_child("sources"),
+          solution_augmented_ensemble[member].block(0), timers);
```
Rombur, line 1369: I don't think you need `temp_time`. This should work:
```suggestion
      time = thermal_physics_ensemble[member]->evolve_one_time_step(
          old_time, time_step, database_ensemble[member].get_child("sources"),
          solution_augmented_ensemble[member].block(0), timers);
```

application/adamantine.hh
```diff
@@ -1314,6 +1383,14 @@ run_ensemble(MPI_Comm const &communicator,
...
     if (assimilate_data)
     {
+      /*
+    // TESTING
+    for (unsigned int member = 0; member < ensemble_size; ++member)
+    {
+      solution_augmented_ensemble[member].collect_sizes();
+    }
+    */
+
```
Rombur, line 1393: Remove this block

application/adamantine.hh
```diff
@@ -1375,10 +1453,34 @@ run_ensemble(MPI_Comm const &communicator,
...
+            }
+            else if (augmented_state_parameters.at(index) ==
+                     adamantine::AugmentedStateParameters::beam_0_max_power)
+            {
+              database_ensemble[member].put(
+                  "sources.beam_0.max_power",
+                  solution_augmented_ensemble[member].block(1)[index]);
+            }
+          }
+        }
```
Rombur, line 1482: Add a `FIXME` that we need to generalized this.

Rombur (commented):
I did a quick pass. I'll take another look later

README.md
stvdwtt: Correct, this only applies to `beam_0`. I agree that it's clunky, but I'm not sure what a better interface is. In theory the max power or the absorption efficiency could differ between beams, so I don't think we want to have one value that applies to all beams (but that would be an easy change). 

Previously I've thought above moving all of the ensemble/data assimilation parameter information to the relevant parameter blocks (e.g. the standard deviation in the max power would be in the general beam 0 parameter block). That's cleaner in some ways, but necessarily means that all the ensemble/data assimilation information is scattered through the input file.

I'm open to any ideas on how to improve this.

application/adamantine.hh
stvdwtt: Good catch -- that's leftover from a previous implementation I tried. I'm removing that.

application/adamantine.hh
stvdwtt: Good idea -- I carried that into DataAssimilator as well.

application/adamantine.hh
stvdwtt: Good idea, that works.

stvdwtt:
I broke the three tests that were inside `test_integration_3d`, so hopefully the tests don't time out anymore.

application/adamantine.hh
```diff
@@ -1030,22 +1039,94 @@ run_ensemble(MPI_Comm const &communicator,
...
+  {
+    // PropertyTreeInput data_assimilation
+    data_assimilation_database = data_assimilation_optional_database.get();
+
+    // PropertyTreeInput data_assimilation.assimilate_data
+    if (data_assimilation_database.get("assimilate_data", false))
+    {
+      assimilate_data = true;
+
+      if (data_assimilation_database.get("augment_with_beam_0_absorption",
```
Rombur, line 1070: Missing `// PropertyTreeInput ...` comment

application/adamantine.hh
```diff
@@ -1030,22 +1039,94 @@ run_ensemble(MPI_Comm const &communicator,
...
+                                         false))
+      {
+        augmented_state_parameters.push_back(
+            adamantine::AugmentedStateParameters::beam_0_absorption);
+      }
+      if (data_assimilation_database.get("augment_with_beam_0_max_power",
```
Rombur, line 1076: Same as above

source/DataAssimilator.cc
```diff
@@ -293,6 +317,23 @@ void DataAssimilator::update_covariance_sparsity_pattern(
...
+    }
+  }
+
+  for (unsigned int i1 = 0; i1 < _sim_size; ++i1)
+  {
+    for (unsigned int j1 = _sim_size; j1 < augmented_state_size; ++j1)
+    {
+      dsp.add(i1, j1);
+    }
+  }
```
Rombur, line 335: I have two questions:
 1. Should this be a regular sparse matrix or a BlockSparseMatrix?
 2. It looks like the new entries are dense not sparse. Everything is coupled together?

source/DataAssimilator.hh
```diff
@@ -36,6 +37,12 @@ enum class LocalizationCutoff
   none
 };
 
+enum class AugmentedStateParameters
+{
+  beam_0_absorption,
+  beam_0_max_power
+};
```
Rombur, line 44: We will need to rework that too but it can wait for another PR

source/ThermalPhysics.templates.hh
```diff
@@ -719,9 +719,27 @@ template <int dim, int fe_degree, typename MemorySpaceType,
...
+        heat_source_database.get_child("beam_" + std::to_string(source_index));
+
+    // PropertyTreeInput sources.beam_X.type
+    std::string type = beam_database.get<std::string>("type");
+
+    if (type == "goldak" || type == "electron_beam")
+      source->set_beam_properties(beam_database);
+
+    source_index++;
+  }
```
Rombur, line 742: Is this the right place to do this? Wouldn't it better to put that in `run_ensemble`?

application/adamantine.hh
stvdwtt: Done.

application/adamantine.hh
stvdwtt: Done.

source/DataAssimilator.cc
stvdwtt: You're right that the new entries are dense and that everything is coupled together. That's because the parameters nominally have a global impact (most apparent for something like the thermal conductivity). There are some methods for localization for parameters, but I don't have a good enough handle on how appropriate they are for us (and how accepted they are). That's definitely something we can keep an eye on. For now, I don't it hurts us much to keep them dense as long as we only have a few parameters augmenting the state at a time.

I think a regular sparse matrix is appropriate because the covariance matrix is only ever used for the entire augmented ensemble. DataAssimilator is almost entirely agnostic to whether the state is augmented or not (with `calc_Hx` being the exception, since the parameters are by definition not observable). The only advantage I can see to a BlockSparseMatrix would be if it could handle the blocked nature of the sparsity pattern better (one big truly sparse block for the base state and then three smaller dense blocks).

source/DataAssimilator.hh
stvdwtt: Agreed. Since reworking the inputs isn't on our immediate to-do list, I'll make an issue for it.

source/ThermalPhysics.templates.hh
Rombur: Yes that sounds good.

stvdwtt:
I think I fixed the issue and now the only problem is that the clang sanitizer tests time out for two tests. I increased the test timeout to 3h. That's much longer than it feels like test should take, but I don't want to modify tests in this PR (especially since one of the tests that times out is the DA test).

Rombur (approved):

Issue #139, opened 2022-01-28
Unexpected thermal evolution with finer meshes


stvdwtt:
I'm seeing an odd bug using the current master branch of the code. For coarse meshes, the code behaves as expected. For somewhat finer meshes, the temperature at the external boundaries is frozen at the initial temperature, but the temperature inside the bulk of the material evolves "normally". For even finer meshes, the entire domain remains frozen at the initial temperature.

This issue appears to be related to the absolute size of the elements, with issues arising once they fall below ~5e-5. Changing between adiabatic and non-adiabatic BCs has no effect.

An example of an input file that triggers this issue is below (with the input file given an extra ".txt" extension so that GitHub lets me attached it). As is, there is evolution in the bulk but the boundary values are frozen. If `length_divisions` and `width_divisions` are changed to 8, the simulation proceeds as normal. If they are changed to 32, then the entire solution is frozen.

[bare_plate_L_scan_path.txt](https://github.com/adamantine-sim/adamantine/files/7961430/bare_plate_L_scan_path.txt)
[bare_plate_unaugmented_ref.info.txt](https://github.com/adamantine-sim/adamantine/files/7961433/bare_plate_unaugmented_ref.info.txt)

Could this be due to misapplied constraints?

Rombur:
Fixed by #140

PR #140, opened 2022-02-01
Remove threshold check for inverting the mass matrix

stvdwtt:
As we hoped, this fixes the problem. Removing the divide by zero check doesn't seem to cause any issues for applying BCs or material deposition.


Rombur (approved):
That's great! Let's merge it once the CI is done

Rombur:
Fixes #139

PR #142, opened 2022-02-24
Fix AMR bugs and add an AMR test

stvdwtt:
The AMR bugs were fairly straightforward to fix. Back from #109, we were missing a call to `get_state_from_material_properties()` in `refine_and_transfer()`. Back from #135 there was a copy/paste error that set `cos` twice instead of `cos` then `sin`.

I also added an AMR integration/regression test.


stvdwtt:
Gold values for the new test accidentally correspond to a different number of time steps that I was trying out. Will push a fix this morning.

stvdwtt:
@Rombur, don't review/merge this yet -- I found another issue. I'm modifying the test to trigger it and then adding a prospective fix.

application/adamantine.hh
```diff
@@ -886,8 +889,12 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
         (activation_start == activation_end) ? false : true;
 #endif
     timers[adamantine::evol_time].start();
+
     time = thermal_physics->evolve_one_time_step(time, time_step, solution,
                                                  timers);
+    // Apply constraints
+    affine_constraints.distribute(solution);
```
stvdwtt, line 896: @Rombur, I'm still not 100% on when constraints need to be distributed. It appears that they need to be distributed here every time step, but I'm not sure if there's a better time/place to do this.

Rombur: Yeah, I am not sure exactly either. Clearly `distribute` needs be called before the mesh is changed and before you output the results. I am not sure it is necessary to do that at every time step but it clearly doesn't hurt. If you do it here,  you should remove the `distribute` in  `output_pvtu()`.

stvdwtt: Earlier I tried it _after_ the mesh was changed, but not before. I'll give before the mesh is changed a try. If that doesn't work, then I'll leave it as is but remove it in the output function.

stvdwtt: You were right, adding it before the mesh is changed is what is necessary. For integration testing purposes, I also had to add it at the very end as well.

stvdwtt:
@Rombur, I just created an issue (#143) for another AMR issue. If the checks pass for this PR, are you ok to merge this as-is and leave the fix for #143 in another PR?

Rombur:
Yes, that's fine.

Rombur (approved):

PR #144, opened 2022-03-08
Use ArborX to speedup localization

Rombur:
deal.II is missing some wrappers for ArborX. There is a bunch of code that will move there in the future. Once that's done and we update deal.II in the CI, I will remove the code in adamantine


stvdwtt (commented):
Looks good other than the typo in the comment.

Rombur:
Fixed

PR #145, opened 2022-04-06
AMR fixes for ensemble calculations

stvdwtt:
This PR makes several changes to ensemble calculations, mostly relating to AMR:
- It resizes the blocks appropriately for the solution vector when the number of DOFs change
- It removes the copy of the constraints in `run_ensemble`, changing to direct access from `ThermalPhysics` to prevent the constraints from being out of sync
- It applies the constraints before data assimilation
- For AMR generally, it changes how the beam intensity is checked, adding checks at face centers
- For augmented state data assimilation, it prints out the augmented parameters before and after each data assimilation operation
- AMR is added to the data assimilation integration test

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -506,11 +506,22 @@ compute_cells_to_refine(
                triangulation.active_cell_iterators(),
                dealii::IteratorFilters::LocallyOwnedCell()))
       {
+        // Check the value at the center of the cell faces and cell center.
+        // Checking both is helpful when starting with a coarse mesh when the
+        // radius of the beam is smaller than the element size.
+        bool refine = false;
+        for (unsigned int f = 0; f < cell->reference_cell().n_faces(); ++f)
+        {
+          if (beam->value(cell->face(f)->center(), current_time,
+                          current_source_height) > refinement_beam_cutoff)
+            refine = true;
+        }
         if (beam->value(cell->center(), current_time, current_source_height) >
             refinement_beam_cutoff)
-        {
+          refine = true;
+
+        if (refine)
           cells_to_refine.push_back(cell);
-        }
       }
     }
   }
@@ -1168,6 +1179,8 @@ run_ensemble(MPI_Comm const &communicator,
       }
     }
 
+    solution_augmented_ensemble[member].collect_sizes();
+
     geometry_ensemble.push_back(std::make_unique<adamantine::Geometry<dim>>(
         communicator, geometry_database));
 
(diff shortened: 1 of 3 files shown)


application/adamantine.hh
```diff
@@ -506,11 +506,22 @@ compute_cells_to_refine(
                triangulation.active_cell_iterators(),
                dealii::IteratorFilters::LocallyOwnedCell()))
       {
+        // Check the value at the center of the cell faces and cell center.
```
Rombur, line 509: There is no point in checking both the faces and the center of cell. There is no way the beam can reach the center of  the cell and miss the faces.

I just want to point out that we could also use the quadrature points inside the cell to decide if the mesh needs to be refine. Not sure it's worth it though.

application/adamantine.hh
```diff
@@ -506,11 +506,22 @@ compute_cells_to_refine(
...
+        // Checking both is helpful when starting with a coarse mesh when the
+        // radius of the beam is smaller than the element size.
+        bool refine = false;
+        for (unsigned int f = 0; f < cell->reference_cell().n_faces(); ++f)
+        {
+          if (beam->value(cell->face(f)->center(), current_time,
+                          current_source_height) > refinement_beam_cutoff)
+            refine = true;
```
Rombur, line 517:
```suggestion
         {
           cells_to_refine.push_back(cell);
           break;
         }
```

application/adamantine.hh
```diff
@@ -1446,10 +1469,26 @@ run_ensemble(MPI_Comm const &communicator,
           std::cout << "Performing data assimilation at time " << time << "..."
                     << std::endl;
 
+        // Print out the augmented parameters
+        if (rank == 0)
```
Rombur, line 1473: We have a verbose flag. You may want to use it too because if the size of the ensemble is 50, it will fill up the screen

application/adamantine.hh
```diff
@@ -1564,6 +1603,19 @@ run_ensemble(MPI_Comm const &communicator,
         if (rank == 0)
           std::cout << "Done." << std::endl;
 
+        // Print out the augmented parameters
+        if (rank == 0)
```
Rombur, line 1607: same as above

Rombur (changes requested):

application/adamantine.hh
Rombur: If you think that won't be a problem, we can let the code as is for now and changed it later if it becomes annoying.

Rombur (approved):

PR #146, opened 2022-04-11
Update CI

Rombur:
This PR updates the Docker images and the CI. There are now two images: a CUDA image and a CPU only image. The new images contains Trilinos, ADOL-C, SymEngine, and dealii-weak_forms. Right now, dealii-weak_forms requires all the backends to be enabled (Sacado, ADOL-C, and SymeEngine). I'll remove them once dealii-weak_forms fixes that problem. I have added a CPU only build in the CI to make sure that everything works when we don't compile deal.II with CUDA support. I don't expect any difference but sometimes the compiler behaves slightly differently.


Rombur:
There was an update in ArborX that breaks the code. I will fix it.

Rombur:
This is ready for review. I have modified one test to make it more robust. We have one test where we do ray tracing and the ray was hitting the edge of the domain. When compiling with `--fast-math` the cell was missed, so I moved the ray to hit the cell further away from the edge. I have added a test using `-O3` and `--fast-math` because we will need all the optimization we can get in production run. In production run, we should also add `-march=native` but this flag can be a problem for the CI because we don't know where it will run.

ci/Dockerfile
```diff
@@ -125,120 +133,115 @@ RUN export P4EST_VERSION=2.3.2 && \
...
+        CFLAGS="-I${HDF5_DIR}/include" \
+        LDFLAGS="-L${HDF5_DIR}/lib -lhdf5" \
+        && \
+    make -j${NPROC} install && \
+    rm -rf ${NETCDF_ARCHIVE} && \
+    rm -rf ${NETCDF_BUILD_DIR} && \
+    rm -rf ${NETCDF_SOURCE_DIR}
+ENV NETCDF_DIR=/opt/netcdf
+
+# We want to enable SEACAS to be able to read EXODUS file but it clashes with ADOL-C. For now,
```
stvdwtt, line 188: Do you think this is medium-term issue that needs to be mentioned in the README, or just a short-term issue that prevents us from testing EXODUS files in the CI, but that perhaps SEACAS and ADOL-C won't clash in other environments/Trilinos versions?

Rombur: IIRC there is clash because both SEACAS and ADOL-C have defined the same macro. Right now what happens is that deal.II detects that both SEACAS and ADOL-C are enabled and the configuration fails. Moving forward, we will only use Sacado and we won't need to install ADOL-C.  Removing the need to enable all the backends is a WIP (see https://github.com/jppelteret/dealii-weak_forms/pull/46). I am hoping that we get this fixed soonish.

stvdwtt (approved):

PR #148, opened 2022-04-19
Clean up documentation

Rombur:
- Clean up README.md
- Fix Doxygen warnings
- Remove Latex file

PR #149, opened 2022-04-20
Add dealii-weak_forms to the build system

Rombur:


stvdwtt (commented):
Is `ADAMANTINE_ENABLE_DEALII_WEAK_FORMS` set to false by default? If so, perhaps it would be worthwhile to set it to true in the CI so that the build process with it is tested.

Also, since this adds a CMake option, I think the README should be updated to include it (either in this PR or #148, if that's cleaner).

Rombur:
I can do it here but I was going to do these two things once we actually use the library. Right now, turning this thing on adds the path to the build system but that's it.

stvdwtt:
Sounds good -- I'll merge it in.

stvdwtt (approved):

PR #155, opened 2022-06-07
Cleanup PR

Rombur:


Rombur:
I think I know what's the problem with the CI not reporting the results. I will fix it but not for this PR. Touching the CI is always a time sink...

PR #160, opened 2022-06-21
Add MechanicalOperator

Rombur:
This adds the core of the `MechanicalOperator` class. There is still a lot of work to do before the class is done but I prefer to get smaller PR into the code. Right now, we can only solve an elastostatic problem, i.e., time-independent elastic problem.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -47,6 +47,17 @@ set(Adamantine_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/validate_input_database.cc
   )
 
+if (ADAMANTINE_ENABLE_DEALII_WEAK_FORMS)
+  set(Adamantine_HEADERS
+      ${Adamantine_HEADERS}
+      ${CMAKE_CURRENT_SOURCE_DIR}/MechanicalOperator.hh
+      )
+  set(Adamantine_SOURCES
+      ${Adamantine_SOURCES}
+      ${CMAKE_CURRENT_SOURCE_DIR}/MechanicalOperator.cc
+     )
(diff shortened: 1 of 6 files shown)


stvdwtt:
@Rombur I'm not sure if you already saw this, but it looks like there's an issue with finding a header file in the test for the mechanical operator.

```
/var/jenkins/workspace/adamantine/tests/test_mechanical_operator.cc:22:10: fatal error: weak_forms/weak_forms.h: No such file or directory
   22 | #include <weak_forms/weak_forms.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
```

ci/Dockerfile
```diff
@@ -25,9 +26,16 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y \
       libpfm4-dev \
       libunwind-dev \
       libgmp-dev \
+      libtbb-dev \
```
Rombur, line 29: This is necessary because otherwise `<execution>` fails to compile. I am thinking that we should remove `<execution>`. It doesn't work with `libc++` or with `libstdc++` without TBB. It's just not very portable and doesn't buy us much.

Rombur:
I cannot get the sanitizer to pass anymore. It dies on the standard library no matter what I do. I propose to disable the sanitizer for now and come back to this later. What do you think @stvdwtt ?

stvdwtt:
@Rombur, I agree that disabling the sanitizer for now is the best path. There's no sense in getting bogged down since it doesn't seem like the issue is a real problem that prevents us from what we need to do with the code.

stvdwtt (approved):

PR #162, opened 2022-06-28
Add MechanicalPhysics

Rombur:


Rombur:
Retest

Rombur:
Retest this please

source/MechanicalPhysics.hh
```diff
@@ -0,0 +1,70 @@
...
+#include <Geometry.hh>
+#include <MechanicalOperator.hh>
+
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/fe/fe_system.h>
+#include <deal.II/hp/fe_collection.h>
+
+namespace adamantine
+{
+// TODO Inherit from Physics or remove Physics entirely?
```
stvdwtt, line 20: I don't have a strong opinion here. I don't know that we're going to have enough other physics classes that it makes sense to force standardization if there isn't substantial shared code between ThermalPhysics and MechanicalPhysics. (Realizing that changing either way is out of scope for this PR.)

Rombur: Yeah I am thinking about removing the Physics class. I think we could reintroduce it if we add another physics model. I want to do some code cleaning once the thermo-mechanics is in a good shape.

Rombur:
OK This looks good. I had to restart the CI a few times because of problems on some on the machines. It was not due to the PR.

stvdwtt (approved):

PR #163, opened 2022-06-28
Use Nearest predicate instead of Spatial predicate to compute the ray-box intersection

Rombur:
This is the bug I talked about this morning. I switched from `Intersect` to `Nearest`. It's a little bit weird but the `BoundingBox` nearest to a `Ray` is the first `BoundingBox` that the `Ray` intersects. If a `Ray` misses a `BoundingBox`, the distance between the `Ray` and the `BoundingBox` is infinite.


source/experimental_data.cc
```diff
@@ -34,17 +34,18 @@
 namespace adamantine
 {
 /**
- * This class implements a predicate for the intersection of Ray and
- * BoundingBox
+ * This class implements a predicate for the intersection of Ray and the first
+ * BoundingBox. In ArborX, this is a nearest search between a Ray and the
+ * BoundingBoces.
```
stvdwtt, line 39: Typo:

```suggestion
 * BoundingBoxes.
```

Rombur:
Retest this please

Rombur:
There are a couple things on this PR:

1. I had to increase the tolerance on the `MechanicalPhysics` test because otherwise I would get some errors. I think it's fine because the test uses CG so, there are some extra round off errors
2. clang-sanitizer is acting once again. I propose to disable the sanitizers for now. I will come back to it later.
3. the reason the PR is failing for one configuration is because of the flag `--fast-math`. When using the flag, ArborX misses a cell. 

I am not sure what to do with this PR. I think 1. and 2. should get merged but I am not sure about this PR itself. We can remove `--fast-math` and have the CI passes or just hold on the PR for now. I do want to understand what is going on in ArborX but I don't know how long it will take (I will do that as part of the Alexa project not the LDRD).

stvdwtt:
@Rombur, my thought is to disable `--fast-math` and merge the whole PR. At this point, I think getting the ray intersections correct is more important than performance (within reason), especially since we need to start exercising the the ray-tracing in the DA workflow.

Rombur:
I think I know what is going on. The ray hits an edge and it is parallel to the face. In the distributed case that face is at the limit between the two domains and because of a small round off error the ray misses the face. If I move the ray slightly to hit the face and not the edge the test passes. In practice, we would need to be very unlucky for this to happen since most rays won't be parallel to a face, hit exactly the edge, and hit just between different domains. I can modified the test or remove the `fast-math`. Note that the current code could potentially have the same problem but it is just not triggered by the test.

stvdwtt:
In that case, I think modifying the test is the best way to go. I agree that the issue of hitting or missing an edge with a parallel ray is enough of an edge case (literally, sorry) that we can ignore it without practical repercussions.

stvdwtt (approved):

PR #165, opened 2022-07-18
Refactor PostProcessor to output result of mechanical simulation

Rombur:
Refactor `PostProcessor`  to be able to output thermal results, mechanical results, or thermo-mechanical results. For now, the mechanical output is limited to displacement and strain. Adding stress is not difficult but it requires to also pass the `MaterialProperty`. I've let that for later.

### DIFF
--- a/source/PostProcessor.cc
+++ b/source/PostProcessor.cc
@@ -1,4 +1,4 @@
-/* Copyright (c) 2016 - 2021, the adamantine authors.
+/* Copyright (c) 2016 - 2022, the adamantine authors.
  *
  * This file is subject to the Modified BSD License and may not be distributed
  * without copyright and license information. Please refer to the file LICENSE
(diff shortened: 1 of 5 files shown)


source/PostProcessor.cc
```diff
@@ -15,13 +15,55 @@
...
 template <int dim>
 PostProcessor<dim>::PostProcessor(MPI_Comm const &communicator,
                                   boost::property_tree::ptree const &database,
                                   dealii::DoFHandler<dim> &dof_handler,
                                   int ensemble_member_index)
-    : _communicator(communicator), _dof_handler(dof_handler)
+    : _communicator(communicator)
 {
+  // This is an internal data we embded in the data base. We don't expose
+  // it to the user
```
stvdwtt, line 54: What does "internal" mean here? First I interpreted it to mean that the "thermal_output" and "mechanical_output" entries are set within the PostProcessor class, but that doesn't seem to be true. Does "internal" mean that it gets set in the application, not the input file?

source/PostProcessor.cc
```diff
@@ -15,13 +15,55 @@
...
 namespace adamantine
 {
+template <int dim>
+StrainPostProcessor<dim>::StrainPostProcessor()
+    : dealii::DataPostprocessorTensor<dim>("strain", dealii::update_gradients)
+{
+}
+
+template <int dim>
+void StrainPostProcessor<dim>::evaluate_vector_field(
```
stvdwtt, line 25: The name makes it seem like this function is more general than it is. I'd recommend changing it to something like `calc_strain_from_displacement`.

Rombur: I don't have a choice in the name of the function. The class is derived from `DataPostprocessorTensor` and I need to override that function for deal.II to do the right thing

stvdwtt: Oh, I see, I didn't catch that the class was derived from a deal.II class.

source/PostProcessor.cc
stvdwtt: I think using the property_tree is perfectly fine, I just wanted to make sure I understood the information flow. I have a suggested edit for the comment below.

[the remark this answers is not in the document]
Rombur: I have changed the comment

stvdwtt (approved):

PR #167, opened 2022-08-23
Add MechanicalOperator and MechanicalPhysics

Rombur:


stvdwtt (commented):
@Rombur, this looks good. My only question is why you're using negative temperatures to switch to plain mechanics rather than passing some sort of flag.

Rombur:
No good reason. It's just one less argument to pass. This not the final version of the interface. Using the initial temperature as `T_0` is fine for now but when we have a more realistic case,  we should use`T_0` for the plate and the solidification temperature for the melted material.

stvdwtt (approved):

PR #168, opened 2022-08-23
Fix timer print-out text

stvdwtt:
The text for the timer for reading in experimental data was added twice to the `timers` vector. This lead to incorrect matching between the timer text and the timer counters.

I also changed the simulation start notification to say whether it is an ensemble simulation or not.


Rombur (approved):

Rombur:
Oops merged too quickly. Can you also change `adamantine.cu`

stvdwtt:
Yes, definitely.

PR #173, opened 2022-09-23
Add Caliper, Adiak, and dealii-weak_forms to the adamantine docker image

Rombur:
- Adding `dealii-weak_forms` to the docker image that we produce on merge to allow people to use the mechanical simulation.
- Adding Caliper and Adiak, so user can send us a `cali` file to know why the code is slow.


Rombur:
I don't know we can't see the report of the CI. It is [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/283/pipeline). I will check it once we are done adding the plasticity.

stvdwtt (approved):

PR #175, opened 2022-09-27
Add analytic solution test for thermoelasticity

stvdwtt:
Add a test using the analytic Eshelby solution and make some minor fixes to MechanicalOperator and MechanicalPhysics.

### DIFF
--- a/source/MechanicalOperator.cc
+++ b/source/MechanicalOperator.cc
@@ -43,9 +43,10 @@ template <int dim, typename MemorySpaceType>
 MechanicalOperator<dim, MemorySpaceType>::MechanicalOperator(
     MPI_Comm const &communicator,
     MaterialProperty<dim, MemorySpaceType> &material_properties,
-    double const initial_temperature)
+    double const initial_temperature, bool include_gravity)
     : _communicator(communicator), _initial_temperature(initial_temperature),
-      _material_properties(material_properties)
+      _material_properties(material_properties),
+      _include_gravity(include_gravity)
 {
 }
 
@@ -169,6 +170,7 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
   dealiiWeakForms::WeakForms::MatrixBasedAssembler<dim> assembler;
 
   std::unique_ptr<dealii::hp::FEValues<dim>> temperature_hp_fe_values;
+
   // If the initial temperature is positive, we solve the thermoelastic problem.
   if (_initial_temperature >= 0.)
   {
@@ -235,7 +237,8 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
         dealiiWeakForms::WeakForms::linear_form(test_grad, expansion_tensor)
             .dV();
   }
-  else
+
+  if (_include_gravity)
   {
     // FIXME the formulation below is more widespread but it doesn't work in
     // dealii-weak_forms yet. Keeping the implementation to use it in the
--- a/source/MechanicalOperator.cc
+++ b/source/MechanicalOperator.cc
@@ -167,28 +167,15 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
 
   // Assemble the bilinear from
   dealiiWeakForms::WeakForms::MatrixBasedAssembler<dim> assembler;
-  // FIXME the formulation below is more widespread but it doesn't work in
-  // dealii-weak_forms yet. Keeping the implementation to use it in the future.
-  // assembler +=
-  //     dealiiWeakForms::WeakForms::bilinear_form(test_grad, lambda + mu,
-  //                                               trial_grad)
-  //         .dV() +
-  //     dealiiWeakForms::WeakForms::bilinear_form(test_grad, mu, trial_grad)
-  //         .delta_IJ()
-  //         .dV() -
-  //     dealiiWeakForms::WeakForms::linear_form(test_val,
-  //     rhs_coeff.value(rhs)).dV();
-  assembler +=
-      dealiiWeakForms::WeakForms::bilinear_form(test_grad, stiffness_tensor,
-                                                trial_grad)
-          .dV() -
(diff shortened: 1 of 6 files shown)


source/MechanicalOperator.cc
```diff
@@ -167,51 +167,51 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
...
-  //         .delta_IJ()
-  //         .dV() -
-  //     dealiiWeakForms::WeakForms::linear_form(test_val,
-  //     rhs_coeff.value(rhs)).dV();
-  assembler +=
-      dealiiWeakForms::WeakForms::bilinear_form(test_grad, stiffness_tensor,
-                                                trial_grad)
-          .dV() -
-      dealiiWeakForms::WeakForms::linear_form(test_val, rhs_coeff.value(rhs))
-          .dV();
```
Rombur, line 186: I am guessing you moved this because the test was failing but I would prefer to keep the code structure as it is currently. Enabling thermal expansion is just adding a term on the right-hand-side and the current code makes this clearer. The issue you had is because I used an arbitrary body force in the rhs for testing purpose. I am working on a PR that replaces the arbitrary body force with gravity but if you just want to turn off the body force for now that's fine. 
In general, should we always have the gravity, never have it, or should we allow to turn it on/off (maybe only for testing)? I was planning to use gravity as the body force because:
 1. what we are doing right now is dumb
 2. I have no idea when it becomes important to take gravity into account

source/MechanicalPhysics.cc
```diff
@@ -68,8 +68,8 @@ void MechanicalPhysics<dim, MemorySpaceType>::setup_dofs(
     dealii::LA::distributed::Vector<double, dealii::MemorySpace::Host> const
         &temperature)
 {
-  setup_dofs();
   _mechanical_operator->update_temperature(thermal_dof_handler, temperature);
+  setup_dofs();
```
Rombur, line 72: I am fine with the change but does it really matters? I don't see how it changes the computation.

tests/test_mechanical_physics.cc
```diff
@@ -240,3 +243,169 @@ BOOST_AUTO_TEST_CASE(elastostatic)
   for (unsigned int i = 0; i < reference_solution.size(); ++i)
     BOOST_TEST(solution[i] == reference_solution[i], tt::tolerance(tolerance));
 }
+
+template <int dim>
+class InitialValueT : public dealii::Function<dim>
+{
+public:
+  virtual double value(const dealii::Point<dim> &p,
```
Rombur, line 251: You don't need `virtual` if you have `override`. It is done like that in deal.II because someone thought it was clearer for C++ beginner that the function was still `virtual`

tests/test_mechanical_physics.cc
```diff
@@ -240,3 +243,169 @@ BOOST_AUTO_TEST_CASE(elastostatic)
...
+  }
+};
+
+namespace utf = boost::unit_test;
+
+/*
+ * This unit test uses the analytic solution for the displacement around a
+ * spherical inclusion as a test case. As a result of the boundary conditions
+ * (fixed on one face) and the need to keep the test computationally
+ * inexpensive, a very loose tolerance has been chosen. Also for simplicity, the
```
Rombur, line 275: That's fine but have you tried on a large domain to check that we get the correct results? At some point, I want to add nightly tests and this test would be a good candidate for that. So please add in the comments the parameters which give an accurate result.

source/MechanicalOperator.cc
stvdwtt: You're right, that was my reasoning.

As far as gravity goes -- for most cases we're going to deal with gravity will be negligible. I can see a few cases where gravity might matter (polymers, very large cantilevered parts), but nothing in the near term. 

I recommend adding an option to turn it on/off in the constructor (default off). Having the ability to add a body force could be useful for testing and it's not much extra code to maintain.

If having it optionally on works for you, I can modify the code in this PR to do that (but leaving the body force as the current arbitrary one).

Rombur: sounds good

source/MechanicalPhysics.cc
stvdwtt: The `MechanicalOperator::reinit` call needs `MechanicalOperator::_thermal_dof_handler` to be set.

source/MechanicalOperator.cc
stvdwtt: I've added the optional addition of the body force.

tests/test_mechanical_physics.cc
stvdwtt: Fixed. (And thanks for explaining why that's the standard way it's done in deal.ii)

tests/test_mechanical_physics.cc
stvdwtt: I tweaked the base test to have an adaptive mesh so it gets to a reasonable solution faster. I also included a comment about how to get to less than 5% error.

source/MechanicalOperator.cc
```diff
@@ -167,51 +168,52 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
...
 
   std::unique_ptr<dealii::hp::FEValues<dim>> temperature_hp_fe_values;
+
   // If the initial temperature is positive, we solve the thermoelastic problem.
   if (_initial_temperature >= 0.)
   {
+    assembler += dealiiWeakForms::WeakForms::bilinear_form(
+                     test_grad, stiffness_tensor, trial_grad)
+                     .dV();
```
Rombur, line 179: I should have been clearer. Move the `bilinear_form` before the `if`. So that the `if(_initial_temperature)` and the ` if (_include_gravity)` only include the `linear_form`, i.e., the right-hand-side.

Rombur (approved):

Rombur:
Looks good. I'll merge once the CI passes

PR #180, opened 2022-10-13
Add control for mechanical BCs

stvdwtt:
Adds the ability to choose which faces have zero Dirichlet mechanical BCs (ie mechanically fixed BCs). The direct motivation is that the face used for the MechanicalPhysics tests is different from what makes the most sense for "real" calculations.

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -908,8 +908,6 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
   if (use_mechanical_physics)
   {
-    // For now assume that only the bottom of the domain is fixed in place
-    std::vector<unsigned int> fixed_faces = {4};
     if (use_thermal_physics)
     {
       // Thermo-mechanical simulation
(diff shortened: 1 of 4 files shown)


Rombur (commented):
It's an improvement over the current implementation but it's still wrong if the mesh is given by a user or if the mesh is not a `hyper_rectangle`.  We need to think how we want to expose that to the user. In the meantime, here is what I think we should: just change `interpolate_boundary_values` and use *four* instead of *zero* (your current code is wrong any way if we want to constrain multiple faces because you close the contraints inside a for loop). Add a comment that it corresponds to the bottom face of a `hyper_rectangle` and add a `FIXME`. You will need to change line `92` of `tests/test_mechanical_physics.cc`.

stvdwtt:
@Rombur, that makes sense, I made those changes. I still need to figure out what's going on with the "elastostatic" test, which is failing. It might just be that the deformation due to the downward body force is much smaller when the bottom surface is constrained and the tolerance and/or error comparison needs to be changed.

Rombur:
You can find the error [here](https://cloud.cees.ornl.gov/jenkins-ci/job/adamantine/306/testReport/junit/projectroot/tests/Test___CPU_on_device___test_mechanical_physics_1/). You need to check but it looks like it's a round-off error.

stvdwtt:
I changed how the solutions were compared, going from BOOST_CHECK to BOOST_CHECK_SMALL. There were small differences between the solutions where the solutions were zero that are related to the finite solver tolerances. The tests now pass on my machine.

Rombur (approved):

Issue #172, opened 2022-09-23
Performance optimizations


Rombur:
On the CPU,  we can significantly improve the performance by changing the following:
 - [x] applying the boundary conditions in a matrix-free manner. The biggest bottleneck in the code is the `std::map::find` used to map quadrature-point quantities to cell quantities. Using deal.II 9.5.0, we can actually compute the boundary conditions using the matrix-free framework.
 - [x] in `add_material` more that two third of the time is spent in `compute_inverse_mass_matrix()`. Here also we should use the matrix-free framework.
 - [x] when computing material properties using a polynomial, we spend a very large amount of time computing the powers of the temperature

PR #204, opened 2022-11-17
Use matrix-free to apply boundary condition

Rombur:
This is a pretty big PR that does the following:
 - compute the boundary in a matrix-free way and remove unnecessary functions/variable.
 - fix a bug introduced by #201. Before that PR, adding material would not require load balancing because the number of cells was unchanged. After that PR, we perform load balancing because the number of dofs has changed. This means that the state ratios can be moved around.
 - update the tests. Some tests needed to be updated because we compute the boundary condition in a slightly different ways. The gold test needed to be redone.

Part of #172


stvdwtt:
@Rombur, is this still a draft, should I review it?

Rombur:
It's still a draft

Rombur:
This is ready for review

source/ThermalOperator.cc
```diff
@@ -574,6 +731,76 @@ void ThermalOperator<dim, fe_degree,
...
+            _face_material_id(face, q)[i] = cell_tria->material_id();
+          }
+        }
+
+    for (unsigned int face = n_inner_faces; face < n_faces; ++face)
+      for (unsigned int q = 0; q < fe_face_eval.n_q_points; ++q)
+        for (unsigned int i = 0;
+             i < _matrix_free.n_active_entries_per_face_batch(face); ++i)
+        {
+          // We get one cells associated with the face
```
stvdwtt, line 784: Typo?

```suggestion
          // We get one cell associated with the face
```

source/ThermalOperator.hh
```diff
@@ -217,22 +228,34 @@ private:
...
-   * local_apply, which is const.
+   * Table of the powder fraction inside cells; mutable so that it can be
+   * changed in cell_local_apply which is const.
    */
   mutable dealii::Table<2, dealii::VectorizedArray<double>> _powder_ratio;
   /**
-   * Table of the material index, mutable so that it can be changed in
-   * local_apply, which is const.
+   * Table of the powder fraction on faces; mutable so that it can be changed in
+   * face_local apply which is const.
```
stvdwtt, line 242: Typo:

```suggestion
   * face_local_apply which is const.
```

source/ThermalOperator.cc
```diff
@@ -544,6 +570,138 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::cell_local_apply(
...
+  // We now have four cases:
+  //  - cell_1 = cell_2 = FE_Q: internal face of the activated domain
+  //  - cell_1/2 = FE_Q and cell_2/1 = FE_Nothing/does not exit: boundary of the
+  //      activated domain
+  //  - cell_1/2 = FE_Nothing and cell_2/1 = does not exit: external boundary of
+  //      the deactivated domain
+  //  - cell_1 = cell_2 = FE_Nothing: internal face of the non-activated domain
+  // Since we only care on the faces that are at the boundary of the activated
+  // domain, we need to check that cell_1 is different than cell_2 and that at
+  // one of the two cells is using FE_Q
```
stvdwtt, line 591: Any changes would be out of the scope of this PR, but I want to understand if this is worth looking at further.

Would you expect slow on the GPU with all these selection statements? Would there be any possibility of pre-calculating interface indices whenever the mesh changes and just loop through those? Or is that not possible for deal.II reasons? I'm not trying to prematurely optimize, but if there's the potential for an alternate way it'd be worth adding a note/issue.

source/ThermalPhysics.templates.hh
```diff
@@ -283,7 +279,7 @@ ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::ThermalPhysics(
 
   // Create the QCollection
   _q_collection.push_back(QuadratureType(fe_degree + 1));
-  _q_collection.push_back(QuadratureType(1));
+  _q_collection.push_back(QuadratureType(fe_degree + 1));
```
stvdwtt, line 282: This is the `FENothing` quadrature? Do they need to have the same `QuadratureType` for the face calculations to work?

source/ThermalPhysics.templates.hh
```diff
@@ -585,32 +581,70 @@ void ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
...
 #endif
 
+  // Update the material state from the ThermalOperator to MaterialProperty
+  // because, for now, we need to use state from MaterialProperty to perform the
+  // transfer to the refined mesh.
+  set_state_to_material_properties();
+
   _thermal_operator->clear();
+  // The data on each cell is stored in the following order: solution, direction
+  // of deposition (cosine and sine), and state rate.
```
stvdwtt, line 591: "...and state rate."
Typo? Numer of states?

stvdwtt (changes requested):

source/ThermalOperator.cc
Rombur: First, right now we cannot loop over the faces on the GPU, it only works on the CPU. I still need to add that capability to deal.II. The way `FE_Nothing` is treated is very different on the CPU and the GPU. On the CPU, there is a lot more support for hp. On the GPU side, there is no support for hp but you can loop over a subdomain using a predicate. When looping on the face, what we are actually doing here is to use the DG infrastructure. That's why we loop over every face in the domain. For the GPU version, I will once again a predicate and so the filtering will be done during setup.

source/ThermalOperator.cc
```diff
@@ -544,6 +570,138 @@ void ThermalOperator<dim, fe_degree, MemorySpaceType>::cell_local_apply(
...
+          (conv_heat_transfer_coef * (temperature - conv_temperature_infty) +
+           rad_heat_transfer_coef * (temperature - rad_temperature_infty));
+      //  for (unsigned int n = 0; n < rad_temperature_infty.size(); ++n)
+      //  {
+      //    std::cout << "aaa val " << boundary_val[n] << " -inv_rho_cp "
+      //              << -inv_rho_cp << " conv_heat " <<
+      //              conv_heat_transfer_coef[n]
+      //              << " T " << temperature[n] << " T_infty "
+      //              << conv_temperature_infty[n] << std::endl;
+      //  }
```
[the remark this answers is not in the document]
Rombur: Yes, I was using it for debugging

source/ThermalPhysics.templates.hh
Rombur: yes unfortunately. The DG matrix-free framework does not fully support hp

source/ThermalPhysics.templates.hh
Rombur: state ratio.

source/ThermalOperator.cc
stvdwtt: Ah, ok. Thanks!

stvdwtt (approved):
This looks good to me, as soon as the CUDA test passes I'll merge it in.

Rombur:
The CUDA test failed because the driver on the machine the CI was running is too old.  I have restarted the CI (see [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/320/pipeline))

stvdwtt:
Oh good, I was wondering how so many of the tests failed. Merging it in now.

PR #206, opened 2023-01-18
Add support for material addition when using GPU.

Rombur:
After this PR is merged, AMR will be the only capability missing on GPU. Some of these capabilities are actually run on the CPU but hopefully we can move them on the GPU later on.

[ci](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/327/pipeline)

### DIFF
--- a/source/ThermalPhysics.templates.hh
+++ b/source/ThermalPhysics.templates.hh
@@ -9,6 +9,9 @@
 #define THERMAL_PHYSICS_TEMPLATES_HH
 
 #include <ThermalOperator.hh>
+
+#include <deal.II/base/index_set.h>
+#include <deal.II/lac/read_write_vector.h>
 #if defined(ADAMANTINE_HAVE_CUDA) && defined(__CUDACC__)
 #include <ThermalOperatorDevice.hh>
 #endif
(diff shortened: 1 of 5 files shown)


source/ThermalPhysics.templates.hh
```diff
@@ -625,7 +635,8 @@ void ThermalPhysics<dim, fe_degree, MemorySpaceType, QuadratureType>::
       if (cell->active_fe_index() == 0)
       {
         std::vector<double> cell_data(direction_data_size + n_material_states);
-        cell->get_dof_values(solution, cell_solution);
+        cell->get_dof_values(rw_solution, cell_solution);
+        // cell->get_dof_values(solution, cell_solution);
```
stvdwtt, line 639: This commented out line can be deleted, right?

stvdwtt (commented):
Other than deleting a commented-out line, this looks good to merge.

source/ThermalPhysics.templates.hh
Rombur: Thanks. Fixed

stvdwtt (approved):

PR #209, opened 2023-01-24
Thermoelasticity with reference temperatures

stvdwtt:
This PR does the following things:
- Adds a new state that tracks if the cell has been melted
- Adds functionality to adjust the state if the cell goes above a certain temperature
- Changes the thermal expansion to be vs a reference temperature that depends on whether the cell has melted
- Tests

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -913,7 +913,8 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
     material_reference_temps.push_back(reference_temperature);
   }
 
-  // Create MechanicalPhysics if necessary
+  // Create MechanicalPhysics if necessary and create PostProcessor
+  std::unique_ptr<adamantine::PostProcessor<dim>> post_processor;
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
   std::unique_ptr<adamantine::MechanicalPhysics<dim, MemorySpaceType>>
       mechanical_physics;
@@ -927,15 +928,24 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
             communicator, fe_degree, geometry, material_properties,
             material_reference_temps);
     post_processor_database.put("mechanical_output", true);
+
+    post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
+        communicator, post_processor_database,
+        thermal_physics->get_dof_handler(),
+        mechanical_physics->get_dof_handler());
+  }
+  else
+  {
+    post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
+        communicator, post_processor_database,
+        thermal_physics->get_dof_handler());
   }
 
-  adamantine::PostProcessor<dim> post_processor(
-      communicator, post_processor_database, thermal_physics->get_dof_handler(),
-      mechanical_physics->get_dof_handler());
 #else
-  adamantine::PostProcessor<dim> post_processor(
+  post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
       communicator, post_processor_database,
       thermal_physics->get_dof_handler());
+
 #endif
 
   dealii::LA::distributed::Vector<double, MemorySpaceType> temperature;
@@ -977,7 +987,7 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   unsigned int n_time_step = 0;
   double time = 0.;
   // Output the initial solution
-  output_pvtu(post_processor, cycle, n_time_step, time, thermal_physics,
+  output_pvtu(*post_processor, cycle, n_time_step, time, thermal_physics,
               temperature,
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
               mechanical_physics,
@@ -1190,7 +1200,7 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
       {
         thermal_physics->set_state_to_material_properties();
       }
-      output_pvtu(post_processor, cycle, n_time_step, time, thermal_physics,
+      output_pvtu(*post_processor, cycle, n_time_step, time, thermal_physics,
                   temperature,
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
                   mechanical_physics,
@@ -1203,7 +1213,7 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   CALI_CXX_MARK_LOOP_END(main_loop_id);
 #endif
 
-  post_processor.write_pvd();
+  post_processor->write_pvd();
 
   // This is only used for integration test
   if constexpr (std::is_same_v<MemorySpaceType, dealii::MemorySpace::Host>)
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -872,6 +872,21 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   double const initial_temperature =
       material_database.get("initial_temperature", 300.);
 
+  // Get the reference temperature(s) for the substrate and deposited
+  // material(s)
+  std::vector<double> material_reference_temps = {initial_temperature};
+  // PropertyTreeInput materials.n_materials
+  double const n_materials = material_database.get<unsigned int>("n_materials");
+  for (unsigned int i = 0; i < n_materials; ++i)
+  {
+    // Use the solidus as the reference temperature, in the future we may want
+    // to use something else
+    // PropertyTreeInput materials.material_n.solidus
+    double const reference_temperature = material_database.get<double>(
+        "material_" + std::to_string(i) + ".solidus");
+    material_reference_temps.push_back(reference_temperature);
+  }
+
   // Create MechanicalPhysics if necessary
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
   std::unique_ptr<adamantine::MechanicalPhysics<dim, MemorySpaceType>>
@@ -884,14 +899,18 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
     mechanical_physics =
         std::make_unique<adamantine::MechanicalPhysics<dim, MemorySpaceType>>(
             communicator, fe_degree, geometry, material_properties,
-            initial_temperature);
+            material_reference_temps);
     post_processor_database.put("mechanical_output", true);
   }
-#endif
 
+  adamantine::PostProcessor<dim> post_processor(
+      communicator, post_processor_database, thermal_physics->get_dof_handler(),
+      mechanical_physics->get_dof_handler());
+#else
   adamantine::PostProcessor<dim> post_processor(
       communicator, post_processor_database,
       thermal_physics->get_dof_handler());
+#endif
 
   dealii::LA::distributed::Vector<double, MemorySpaceType> temperature;
   dealii::LA::distributed::Vector<double, dealii::MemorySpace::Host>
@@ -1049,6 +1068,20 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
     }
     timers[adamantine::add_material_activate].stop();
 
+    // If thermomechanics are being solved, mark cells that are above the
+    // solidus as cells that should have their reference temperature reset. This
+    // cannot be in the thermomechanics solve because some cells may go above
+    // the solidus and then back below the solidus in the time between
+    // thermomechanical solves.
+    if (use_thermal_physics && use_mechanical_physics)
+    {
+      // FIXME: Need to figure out the new implementation
+      /*
+        thermal_physics->mark_cells_above_temperature(material_reference_temps[1],
+                                                      temperature);
+      */
+    }
+
     // Time can be different than time + time_step if an embedded scheme is
     // used. Note that this is a problem when adding material because it
     // means that the amount of material that needs to be added is not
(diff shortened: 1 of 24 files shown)


stvdwtt:
The [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/332/pipeline) failed, I'm going to look into it this afternoon. Oddly, it's passing on my machine. It could be an inconsistency between test input files in the repo and in my build directory.

Rombur:
I don't know if you know but if you click on the `Tests` button on the top, you can see the error messages from the tests (see [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/332/tests)). I don't think the problem is the input file. Are you using the latest version of the Docker image?

stvdwtt:
I'm using the lastest version of `rombur/adamantine-stack:latest`, do you know if that has diverged from `rombur/adamantine:latest`?

Rombur:
We use `rombur/adamantine-stack:latest` for the CI not `rombur/adamantine:latest`. `rombut/adamantine:latest` is automatically created when we merge a PR.

stvdwtt:
I added a few more changes to address the errors in the CI. 

- Fixes a mistake in how I handled hp elements
- Changes the PostProcess object to a unique pointer to address cases where mechanics is and is not included
- Changes Jenkins input to compile the "CPU" test with deal.ii weak forms
- Adds a check that deal.ii weak forms is linked if the input says that mechanical physics should be active (addresses #210)

Link to the new CI: https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/338/pipeline/

stvdwtt:
@Rombur, the CI passed and this is ready for review.

application/adamantine.hh
```diff
@@ -882,27 +898,56 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
   }
-#endif
 
-  adamantine::PostProcessor<dim> post_processor(
+#else
+  post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
       communicator, post_processor_database,
       thermal_physics->get_dof_handler());
 
+#endif
```
Rombur, line 949: Let's not do this because we still need to add  a case where only the mechanical simulation is enabled. It's also harder to read because of the mix of `#ifdef` and `if`. Instead let's do sth like this:
```c++
thermal_output = post_processor_database.get("thermal_output", false);
mechanical_output = post_processor_database.get("mechanical_output", false);
if ((thermal_output) && (mechanical_output))
  {
     post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
        communicator, post_processor_database,
        thermal_physics->get_dof_handler(),
        mechanical_physics->get_dof_handler());
  }
else
 {
    post_processor = std::make_unique<adamantine::PostProcessor<dim>>(
        communicator, post_processor_database,
        thermal_output ? thermal_physics->get_dof_handler() : mechanical_physics->get_dof_handler());
 }

```

application/adamantine.hh
```diff
@@ -1059,6 +1109,17 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
     timers[adamantine::add_material_activate].stop();
 
+    // If thermomechanics are being solved, mark cells that are above the
+    // solidus as cells that should have their reference temperature reset. This
+    // cannot be in the thermomechanics solve because some cells may go above
+    // the solidus and then back below the solidus in the time between
+    // thermomechanical solves.
+    if (use_thermal_physics && use_mechanical_physics)
+    {
+      thermal_physics->mark_cells_above_temperature(material_reference_temps[1],
```
Rombur, line 1119: Maybe call this `mark_has_melted`? Above temperature is not very obvious which temperature we are talking about.

source/MechanicalOperator.cc
```diff
@@ -84,10 +84,12 @@ template <int dim, typename MemorySpaceType>
 void MechanicalOperator<dim, MemorySpaceType>::update_temperature(
     dealii::DoFHandler<dim> const &thermal_dof_handler,
     dealii::LA::distributed::Vector<double, dealii::MemorySpace::Host> const
-        &temperature)
+        &temperature,
+    std::vector<double> const &has_melted_indicator)
```
Rombur, line 88: Why use a `double`? I think a `bool` would be more appropriate.

source/MechanicalOperator.cc
```diff
@@ -188,12 +190,35 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
...
+          // is not in the unmelted substrate, the reference temperature depends
+          // on the material.
+
+          // FIXME: Needs new implementation
+          bool is_unmelted_substrate = false;
+          if (_has_melted_indicator.at(cell->active_cell_index()) < 0.5)
+          {
+            is_unmelted_substrate = true;
+          }
+          // const bool is_unmelted_substrate = cell->user_flag_set();
```
Rombur, line 203: I guess we don't need this anymore?

source/MechanicalOperator.cc
```diff
@@ -188,12 +190,35 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
           // quadrature point using the temperature DoFHandler.
           auto const &cell = fe_values.get_cell();
 
+          // Get the appropriate reference temperature for the cell. If the cell
+          // is not in the unmelted substrate, the reference temperature depends
+          // on the material.
+
+          // FIXME: Needs new implementation
+          bool is_unmelted_substrate = false;
+          if (_has_melted_indicator.at(cell->active_cell_index()) < 0.5)
```
Rombur, line 199: Why are using `at`? It is significantly slower than `[]`

source/MechanicalOperator.hh
```diff
@@ -102,11 +103,13 @@ private:
    */
   bool _include_gravity = false;
   /**
-   * Initial temperature of the material. If the temperature is positive, we
-   * solve a themo-mechanical problem. Otherwise, we solve a mechanical only
-   * problem.
+   * List of initial temperatures of the material. If the length of the vector
+   * is nonzero, we solve a themo-mechanical problem. Otherwise, we solve a
```
Rombur, line 107:
```suggestion
   * is nonzero, we solve a thermo-mechanical problem. Otherwise, we solve a
```

source/MechanicalOperator.cc
```diff
@@ -188,12 +190,35 @@ void MechanicalOperator<dim, MemorySpaceType>::assemble_system()
...
+
+          double reference_temperature;
+          if (is_unmelted_substrate)
+          {
+            reference_temperature = _reference_temperatures.at(0);
+          }
+          else
+          {
+            reference_temperature =
+                _reference_temperatures.at(cell->material_id() + 1);
```
Rombur, line 213: Why the `+1`?

source/MechanicalOperator.hh
```diff
@@ -141,6 +144,13 @@ private:
    */
   dealii::LA::distributed::Vector<double, dealii::MemorySpace::Host>
       _temperature;
+  /**
+   * Indicator variable for whether a point has ever been above the solidus. The
+   * value is 0 for material that has not yet melted and 1 for material that has
+   * melted. Due to how the state is transferred for remeshing, this has to be
+   * stored as a double.
+   */
```
Rombur, line 152: You can just use a `double` when you refine and then convert back to a `bool`.

source/ThermalPhysics.templates.hh
```diff
@@ -567,6 +569,62 @@ void ThermalPhysics<dim, fe_degree, MemorySpaceType,
...
+      cell_temperature /= cell_volume;
+
+      // Set the indicator that this cell has melted
+      if (cell_temperature > threshold_temperature)
+      {
+        _has_melted_indicator.at(cell->active_cell_index()) = 1.0;
+      }
+    }
+
+  } // namespace adamantine
```
Rombur, line 625: I don't think this comment should be here.

source/validate_input_database.cc
```diff
@@ -24,6 +24,13 @@ void validate_input_database(boost::property_tree::ptree &database)
   bool const use_mechanical_physics = database.get<bool>("physics.mechanical");
   ASSERT_THROW(use_thermal_physics || use_mechanical_physics,
                "Error: both thermal and mechanical physics are disabled");
+  if (use_mechanical_physics)
+  {
+#ifndef ADAMANTINE_WITH_DEALII_WEAK_FORMS
```
Rombur, line 29: Move the `#ifndef` about the `if`

ci/jenkins_config
```diff
@@ -56,8 +56,10 @@ pipeline {
                     cmake \
                     -D CMAKE_BUILD_TYPE=Debug \
                     -D ADAMANTINE_ENABLE_TESTS=ON \
+                    -D ADAMANTINE_ENABLE_DEALII_WEAK_FORMS=ON \
```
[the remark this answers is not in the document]
stvdwtt: Oh, I see. In that case, what should I be doing for the thermoelastic integration test? For obvious reasons it fails without deal.II weak forms. Should the test be catching an exception in the case that deal.II weak forms isn't linked? I'm a little puzzled how the mechancial_operator and mechanical_physics tests were passing.

application/adamantine.hh
```diff
@@ -882,27 +898,56 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
+  std::unique_ptr<adamantine::PostProcessor<dim>> post_processor;
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
   std::unique_ptr<adamantine::MechanicalPhysics<dim, MemorySpaceType>>
       mechanical_physics;
   if (use_mechanical_physics)
   {
     // PropertyTreeInput discretization.mechanical.fe_degree
     unsigned int const fe_degree =
-        discretization_database.get<unsigned int>("mechanial.fe_degree");
+        discretization_database.get<unsigned int>("mechanical.fe_degree");
```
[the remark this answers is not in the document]
stvdwtt: I don't think it did. We didn't have any integration tests with mechanics.

application/adamantine.hh
```diff
@@ -414,13 +418,17 @@ void refine_and_transfer(
             thermal_physics->get_deposition_cos(activated_cell_id);
         cell_data[n_material_states + 1] =
             thermal_physics->get_deposition_sin(activated_cell_id);
+        cell_data[n_material_states + direction_data_size] =
+            thermal_physics->get_has_melted_indicator(activated_cell_id);
```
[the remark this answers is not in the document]
stvdwtt: Good idea, I'll make that change.

application/adamantine.hh
stvdwtt: Sounds good -- that fits with everything else better.

source/MechanicalOperator.cc
stvdwtt: I agree that a `bool` would fit better, but my understanding is that it has to be a `double` to get packed into the state. It's a bit more indirect, but if you think it would be better, I can store the vector as `bool`s and then convert to `double`s element by element as I pack and unpack the state.

source/MechanicalOperator.cc
stvdwtt: Right, we don't use the `user_flag`s anymore

ci/jenkins_config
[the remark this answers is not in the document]
Rombur: The tests are not run if we don't have dealii-weak_forms. See [here](https://github.com/adamantine-sim/adamantine/blob/master/tests/CMakeLists.txt#L69-L73)

source/MechanicalOperator.cc
stvdwtt: I'm switching it to `[]`, that was to help me catch some indexing issues.

source/MechanicalOperator.cc
stvdwtt: The reference temperature vector looks like: {initial condition temperature, solidus material 0, solidus material 1, ...}

source/MechanicalOperator.hh
stvdwtt: That makes sense, I'll make that change.

source/ThermalPhysics.templates.hh
stvdwtt: Good catch, removed.

ci/jenkins_config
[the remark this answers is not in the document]
stvdwtt: Ah, I see. Thanks! I'll move that test to the WEAK_FORMS_UNIT_TESTS and revert the CPU cmake options.

source/MechanicalOperator.cc
Rombur: What about we switch the order? {solidus_material_0, solidus material, ..., initial condition temperature} Then when you call `_reference_temperature[cell->material_id()]`, you get what you would expect.

stvdwtt: I like that idea a lot

application/adamantine.hh
```diff
@@ -882,26 +907,59 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
+    // to use something else
+    // PropertyTreeInput materials.material_n.solidus
+    double const reference_temperature = material_database.get<double>(
+        "material_" + std::to_string(i) + ".solidus");
+    material_reference_temps.push_back(reference_temperature);
+  }
+  material_reference_temps.push_back(initial_temperature);
+
+// Create MechanicalPhysics
 #ifdef ADAMANTINE_WITH_DEALII_WEAK_FORMS
```
Rombur, line 928: This needs to be moved down after `mechanical_physics` so that `mechanical_physics` always exist because now we use it lower.

application/adamantine.hh
```diff
@@ -518,6 +532,14 @@ void refine_and_transfer(
             transferred_data[total_cell_id][n_material_states]);
         transferred_sin.push_back(
             transferred_data[total_cell_id][n_material_states + 1]);
+
+        // Convert from double back to bool
+        if (transferred_data[total_cell_id]
+                            [n_material_states + direction_data_size] > 0.5)
+          has_melted.push_back(true);
+        else
+
```
Rombur, line 541: There is another extra blank line

stvdwtt:
I added the dummy `MechanicalPhysics` and I believe have addressed all of the review comments.

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/343/pipeline)

Rombur (approved):

Issue #187, opened 2022-10-17
Add support for FE_Nothing for mechanics


stvdwtt:
This will enable mechanics to work with material deposition.

PR #214, opened 2023-01-31
Use FE_Nothing for everything but solid for mechanical simulation

### LINKED ISSUE #185: Decide how to handle liquid for mechanics calculations
Options:
1. FE_Nothing for liquid, only do calculation for the solid
2. Set different mechanical properties in the liquid (essentially treating it as a very deformable solid)
3. Continue treating the liquid as a solid with the same mechanical properties as the solid phase

Rombur:
Changes are trivial because we already supported `FE_Nothing`. I just had to tag the cells that are not solid with `FE_Nothing`.

Closes #185 and #187

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -1,4 +1,4 @@
-/* Copyright (c) 2016 - 2022, the adamantine authors.
+/* Copyright (c) 2016 - 2023, the adamantine authors.
  *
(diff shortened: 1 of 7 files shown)


stvdwtt (approved):

stvdwtt (approved):
I'll merge this in now, but eventually we'll want to add an integration test that covers the case of when thermal physics has FE_Nothing.

PR #217, opened 2023-01-31
Fix bug with ghosts zeroed out after adding material

### LINKED ISSUE #212: Zeros in the temperature introduced along MPI boundaries after adding material
Every time that material is added in an MPI run, zeros are introduced along the MPI domain boundaries. At first I assumed this was a ghost cell issue, but adding a simple `temperature.update_ghost_values()` after the call to `add_material` doesn't do anything. I'm not sure when this started happening.

@Rombur, I can keep digging, but if it's obvious to you what's wrong, let me know.

stvdwtt:
Fixes #212 by switching from using `dealii::VectorOperation::min` to `dealii::VectorOperation::insert`, as in:
https://github.com/dealii/dealii/pull/14340

This PR also changes the 3D integration test to an MPI test to catch a similar issue in the future.

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/347/pipeline)


tests/test_integration_3d.cc
```diff
@@ -32,11 +32,62 @@ BOOST_AUTO_TEST_CASE(integration_3D, *utf::tolerance(0.1))
...
+  double min_expected = 285.0;
+
+  if (num_ranks == 1)
+  {
+    std::ifstream gold_file("integration_3d_gold.txt");
+
+    for (unsigned int i = 0; i < temperature.locally_owned_size(); ++i)
+    {
+      BOOST_CHECK(temperature.local_element(i) > min_expected);
+      BOOST_CHECK(temperature.local_element(i) < max_expected);
```
Rombur, line 49: Why do you have this check? It should already be taken care with the comparison with the gold value.

Rombur (approved):

Rombur:
Looks like it there is a problem https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/347/tests

Rombur:
We use `signaling_NaN()` only in `MaterialProperties` so it's probably not coming from there.

stvdwtt:
It's looking like the issue with `test_material_deposition` isn't related to adding material. The solution is all NANs after the first `evolve_one_time_step` call is made. I'm looking for the cause of that now.

stvdwtt:
It turns out that `test_material_deposition` was using radiative BCs but didn't have the appropriate material parameters defined. I'm guessing that the solution has been NANs all along, but it wasn't until the `add_material` fix that it was exposed. I changed the BCs to adiabatic.

stvdwtt:
@Rombur, in the most recent CI run, `test_thermal_operator_1` failed in Release. Nothing changed relevant to that test since my last push. I've occasionally seen this happen on my laptop, where randomly this test fails and then passes on a rerun.

Should we just loosen the tolerance? The values are all close. Is there a chance that something is actually wrong?

Rombur:
I don't want to loosen the tolerance right now because I would like to see how often this happens. In release mode, I have compiled deal.II with `-Ofast` which allows the compiler more freedom to reorder some mathematical operations. Before we loosen the tolerance, I would like to see if switching to `-O3` fixes the problem. It's also possible that we read an uninitialize value somewhere. We should try to use `signaling_NaN` more often in debug mode that's the easiest way to find this kind of bug until we get the sanitizers back. For now, I'll restart the CI.

Rombur (approved):

PR #218, opened 2023-02-09
Fix issue when adding material for thermo-mechanical simulation

### LINKED ISSUE #187: Add support for FE_Nothing for mechanics

Rombur:
Fix #187


Rombur:
Somehow this PR triggers a bug here https://github.com/adamantine-sim/adamantine/blob/1f3dbfe963d0a2913354c35bedcac9b0205b6c79/source/ThermalPhysics.templates.hh#L594

Using `at` crashes the code

Rombur:
The code  above is wrong. It won't work with MPI because `cell->active_cell_index()` takes into account ghost and artificial cells.

Rombur:
It's also wrong with `FE_Nothing` because `_has_melted` has the size of the number of active cells which have fe index = 0 which is less than the number of cells when there is `FE_Nothing`.

Rombur:
The tests are now passing [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/352/pipeline)

I also have a clean results using `ctest -T memcheck` which uses valgrind and it looks clean (only leaks from MPI)

stvdwtt (approved):

Rombur:
> There is a bug here
> 
> https://github.com/adamantine-sim/adamantine/blob/1f3dbfe963d0a2913354c35bedcac9b0205b6c79/source/ThermalPhysics.templates.hh#L594
> 
> Using `at` crashes the code. I don't think this is related to this PR

Actually we have the same bug here:
https://github.com/adamantine-sim/adamantine/blob/cb20d62c8c37a47d920bc5f34f668e7878ac59aa/source/MechanicalOperator.cc#L200

This is going to be harder to fix because we don't know in which order the cells are visited

Issue #181, opened 2022-10-17
Harden reading IR data from file


stvdwtt:

stvdwtt:
To close this out I'd like to add the ability to visualize the experimental data on the mesh. That might be via an option in the input file that creates special output files for the experimental data.

stvdwtt:
Addressed in #236

PR #219, opened 2023-02-10
Fix experimental data with FE_Nothing, refactor experimental data interface

stvdwtt:
Partially address #181.

This PR makes the extraction of the DOFs with experimental data work when some elements are FE_Nothing. It also refactors the interface between the functions in `experimental_data` and the `DataAssimilator`. The pair of vectors that map between the experimental observation indices and the DOF indices is now calculated in `experimental_data`, instead of an intermediate being passed through `run_ensemble`. The updated interface makes the "dummy" temperature field in `run_ensemble` unnecessary.

The test for experimental data with FE_Nothing is real IR data (albeit not particularly well calibrated IR data). The test has a runtime on the longer side, which is due to the time calculating the intersections with the mesh. The DOF has limited impact on the runtime, it scales as the number of IR pixels. We could consider a smaller experimental data file. @Rombur what do you think?

### DIFF
--- a/source/experimental_data.cc
+++ b/source/experimental_data.cc
@@ -224,27 +224,13 @@ get_dof_to_support_mapping(dealii::DoFHandler<dim> const &dof_handler)
 
   // Manually do what dealii::DoFTools::map_dofs_to_support_points does, since
   // that doesn't currently work with FE_Nothing
-  const dealii::hp::FECollection<dim> &fe_collection =
-      dof_handler.get_fe_collection();
+  const dealii::FiniteElement<dim> &fe = dof_handler.get_fe(0);
 
-  dealii::hp::QCollection<dim> q_coll_dummy;
-  for (unsigned int fe_index = 0; fe_index < fe_collection.size(); ++fe_index)
-  {
-    q_coll_dummy.push_back(dealii::Quadrature<dim>(
-        fe_collection[fe_index].get_unit_support_points()));
-  }
+  dealii::FEValues<dim, dim> fe_values(fe, fe.get_unit_support_points(),
+                                       dealii::update_quadrature_points);
 
-  dealii::hp::MappingCollection<dim, dim> mapping_collection;
-  const dealii::StaticMappingQ1<dim> mapping;
-  mapping_collection.push_back(mapping.mapping);
-  if (fe_collection.size() > 1)
-    mapping_collection.push_back(mapping.mapping);
-
-  dealii::hp::FEValues<dim, dim> hp_fe_values(mapping_collection, fe_collection,
-                                              q_coll_dummy,
-                                              dealii::update_quadrature_points);
-
-  std::vector<dealii::types::global_dof_index> local_dof_indices;
+  std::vector<dealii::types::global_dof_index> local_dof_indices(
+      fe.n_dofs_per_cell());
 
   for (auto const &cell : dealii::filter_iterators(
            dof_handler.active_cell_iterators(),
@@ -253,16 +239,11 @@ get_dof_to_support_mapping(dealii::DoFHandler<dim> const &dof_handler)
     // only work on locally relevant cells
     if (cell->is_artificial() == false)
     {
-      hp_fe_values.reinit(cell);
-      const dealii::FEValues<dim, dim> &fe_values =
-          hp_fe_values.get_present_fe_values();
-
-      local_dof_indices.resize(cell->get_fe().n_dofs_per_cell());
+      fe_values.reinit(cell);
       cell->get_dof_indices(local_dof_indices);
-
       const std::vector<dealii::Point<dim>> &points =
           fe_values.get_quadrature_points();
-      for (unsigned int i = 0; i < cell->get_fe().n_dofs_per_cell(); ++i)
+      for (unsigned int i = 0; i < fe.n_dofs_per_cell(); ++i)
       {
         indices_points[local_dof_indices[i]] = points[i];
       }


stvdwtt:
CI[](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/353/pipeline)

The CI failed, I'm looking into it.

Rombur:
In ArborX, the rays are defined by an origin and a direction. The direction is normalized. The error message says that the norm of the direction is not greater than zero

Rombur:
> The test has a runtime on the longer side, which is due to the time calculating the intersections with the mesh.

Is it the time spent in ArborX or the time spent on compute the exact intersection in a cell?

stvdwtt:
The error message was related to improper parsing of the header for the CSV file. I modified the README to explicitly state that the file should have a header and now the parser skips the header.

Now the test is running substantially faster (1.33 s on my laptop, 22.91 s on a cloud VM). Maybe the slow performance was related to the uncaught vector magnitude error.

stvdwtt:
New [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/354/pipeline/12)

source/experimental_data.cc
```diff
@@ -212,16 +214,61 @@ std::vector<PointsValues<dim>> read_experimental_data_point_cloud(
...
+
+  dealii::hp::MappingCollection<dim, dim> mapping_collection;
+  const dealii::StaticMappingQ1<dim> mapping;
+  mapping_collection.push_back(mapping.mapping);
+  if (fe_collection.size() > 1)
+    mapping_collection.push_back(mapping.mapping);
+
+  dealii::hp::FEValues<dim, dim> hp_fe_values(mapping_collection, fe_collection,
+                                              q_coll_dummy,
+                                              dealii::update_quadrature_points);
```
Rombur, line 245: All of this can be simplified because we know that we only care about `fe_index=0`.  So you don't need to create an `hp::FEValues` object, you can create a simple `FEValues` object. This also simplify the for loop below.

source/experimental_data.cc
```diff
@@ -212,16 +214,61 @@ std::vector<PointsValues<dim>> read_experimental_data_point_cloud(
...
+
+  std::vector<dealii::types::global_dof_index> local_dof_indices;
+
+  for (auto const &cell : dealii::filter_iterators(
+           dof_handler.active_cell_iterators(),
+           dealii::IteratorFilters::ActiveFEIndexEqualTo(0)))
+  {
+    // only work on locally relevant cells
```
Rombur, line 253: Is this what we really want? I would think that we only care about the locally owned cells.

Rombur: Hmm the only place where you could potentially have a problem is when you do the `compress(VectorOperation::insert)`. I always forget what happens when you insert a local value and do nothing for the equivalent ghost entry. Maybe they need to match and so you need the ghost values?

source/experimental_data.cc
```diff
@@ -224,6 +224,34 @@ get_dof_to_support_mapping(dealii::DoFHandler<dim> const &dof_handler)
...
+    if (cell->is_artificial() == false)
+    {
+      fe_values.reinit(cell);
+
+      local_dof_indices.resize(cell->get_fe().n_dofs_per_cell());
```
Rombur, line 243: This can  be moved outside the loop.

source/experimental_data.cc
```diff
@@ -268,6 +296,7 @@ get_dof_to_support_mapping(dealii::DoFHandler<dim> const &dof_handler)
       }
     }
   }
+  */
```
Rombur, line 299: Did you forget to remove this code?

stvdwtt: Oh yes, my editor crashed after I deleted that and I didn't realize that wasn't saved.

stvdwtt:
New [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/356/pipeline)

source/experimental_data.cc
```diff
@@ -212,16 +214,42 @@ std::vector<PointsValues<dim>> read_experimental_data_point_cloud(
...
+           dealii::IteratorFilters::ActiveFEIndexEqualTo(0)))
+  {
+    // only work on locally relevant cells
+    if (cell->is_artificial() == false)
+    {
+      fe_values.reinit(cell);
+      cell->get_dof_indices(local_dof_indices);
+      const std::vector<dealii::Point<dim>> &points =
+          fe_values.get_quadrature_points();
+      for (unsigned int i = 0; i < cell->get_fe().n_dofs_per_cell(); ++i)
```
Rombur, line 246: You don't need to go through the cell. just above you did `fe.n_dofs_per_cell()`

stvdwtt: Nice catch. I'm about to push that fix after a local test.

Also, the CI failed for `test_thermal_operator_1` for CUDA, which I believe is unrelated to this PR and is that nondeterministic failure again.

stvdwtt: Updated [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/358/pipeline).

Rombur:
I've restarted the PR 3 times but it keeps failing the CI. I think it's related to https://github.com/adamantine-sim/adamantine/pull/218#issuecomment-1428673284 I found the problem using valgrind outside of CTest.

stvdwtt:
@Rombur are you comfortable merging this PR since the test failures are (in all likelihood) unrelated to these changes? 

Confusingly, some of the errors are from CUDA/`test_thermal_operator_1` and some are CPU-on-device/'test_integration_thermoelastic_1`.

Rombur:
Let's merge this. I'll fix the other bug in a different PR

PR #222, opened 2023-02-14
Only loop over locally owned cells in get_dof_to_support_mapping

Rombur:
There is no reason to loop over the ghost cells. On top of that the current code does not work on the MPI case with artificial cell because you cannot check the fe index on an artificial cell. One test needs to be changed because it won't pass with two processors anymore. However, it was passing just by chance. We have one processor that owns the four cells on the bottom of the domain and one processor that owns the four cells on the top and so the ghost + locally owned cells represent the entire mesh. So we are basically solving the same problem as the serial one.

### DIFF
--- a/tests/test_experimental_data.cc
+++ b/tests/test_experimental_data.cc
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, the adamantine authors.
+/* Copyright (c) 2021-2023, the adamantine authors.
  *
  * This file is subject to the Modified BSD License and may not be distributed
  * without copyright and license information. Please refer to the file LICENSE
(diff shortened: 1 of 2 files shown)


tests/test_experimental_data.cc
```diff
@@ -60,61 +60,65 @@ BOOST_AUTO_TEST_CASE(set_vector_with_experimental_data_point_cloud)
...
-  points_values.points.emplace_back(0, 1., 1.);
-  points_values.points.emplace_back(0.5, 0., 1.);
-  points_values.points.emplace_back(0.5, 0.5, 1.);
-  points_values.points.emplace_back(0.5, 1., 1.);
-  points_values.points.emplace_back(1., 0., 1.);
-  points_values.points.emplace_back(1., 0.5, 1.);
-  points_values.points.emplace_back(1., 1., 1.);
-  // Create the values
-  points_values.values = {1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2.};
+  if (dealii::Utilities::MPI::n_mpi_processes(communicator) == 0)
```
stvdwtt, line 63: Shouldn't this be `1` instead of `0`?

Rombur: Good catch

Rombur:
I have added a commit that simplifies the code a little bit

stvdwtt (approved):

PR #228, opened 2023-03-07
Treat experimental data on-the-fly

Rombur:
This PR totally refactors how we deal with experimental. Instead of reading all the data before the simulation, we read the data from a single frame when the simulation reach the time matches the experimental data. If the data is not present yet, we just wait for hit. The code in `experimental_data` has been split into three: `RayTracing`, `PointCloud`, and `experimental_data_utils`. The last file contains all the free functions. [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/372/pipeline)


source/PointCloud.cc
```diff
@@ -0,0 +1,115 @@
...
+                                               std::to_string(camera_id))),
+                           frame_regex, std::to_string(_next_frame));
+    unsigned int counter = 1;
+    while (!boost::filesystem::exists(filename))
+    {
+      // Spin loop waiting for the file to appear
+      if (counter == 0)
+        std::cout << "Waiting for the next frame" << std::endl;
+      ++counter;
+    }
```
stvdwtt, line 56: Should the counter be initialized to 0 here instead of 1? I assume the purpose of this is to print the message once and then quietly increment until the file exists?

source/RayTracing.cc
```diff
@@ -0,0 +1,345 @@
...
+                                               std::to_string(camera_id))),
+                           frame_regex, std::to_string(_next_frame));
+    unsigned int counter = 1;
+    while (!boost::filesystem::exists(filename))
+    {
+      // Spin loop waiting for the file to appear
+      if (counter == 0)
+        std::cout << "Waiting for the next frame" << std::endl;
+      ++counter;
+    }
```
stvdwtt, line 118: Same comment as for PointCloud.

source/PointCloud.cc
```diff
@@ -0,0 +1,115 @@
...
+
+  return _next_frame++;
+}
+
+template <int dim>
+PointsValues<dim>
+PointCloud<dim>::get_points_values(dealii::DoFHandler<dim> const &)
+{
+  return _points_values_current_frame;
+}
```
stvdwtt, line 111: It's a little awkward that the shared interface dictates that the DofHandler is an input here since it isn't needed. That said, there's value in a standardized interface and the only solution I can think of is to add another function in RayTracing that does the actual calculation with the DoFHandler as an input which then turns `get_points_values` into a single "get" function there too.

stvdwtt (commented):
I know there isn't a clean way to include a check in the tests the case where the code has to wait for an experimental data file, but have you checked that manually?

source/PointCloud.cc
Rombur: I don't want to see this message. I've put it there because it's a potentially infinite loop. If this message appears, I can check that the frame really does not exist. It's more of a debugging tool.

source/PointCloud.cc
Rombur: We could change the constructor of `RayTracing` to keep a pointer to the `DoFHandler` if you prefer.

source/PointCloud.cc
stvdwtt: Just to make sure I'm understanding -- the idea is that the message appears only if the file doesn't exist for long enough for `counter` to overflow and make it back to zero?

Rombur: Yes, exactly

stvdwtt: Would you mind changing the comment to read:
```suggestion
    unsigned int counter = 1;
    while (!boost::filesystem::exists(filename))
    {
      // Spin loop waiting for the file to appear (message printed if counter overflows)
      if (counter == 0)
        std::cout << "Waiting for the next frame" << std::endl;
      ++counter;
    }
```

Rombur:
New [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/373/pipeline)

stvdwtt (approved):

Issue #225, opened 2023-02-22
Ray intersection robustness


stvdwtt:
Under some circumstances, when the ray from experimental data is aligned with the mesh, `RayTracing::get_intersection` can erroneously give DOF index 0 as the intersection point. 

There's a note in the function that rays that cross an edge may be lost -- I think this is separate from that, although perhaps related. In the cases I saw, the ray was going through a part of the face away from the edge (and not the exact middle of the face either).

When I have a chance, I'll post a reproducer.

Rombur:
Looking at the code, there are several problem with the current function:
 1. This code below needs to use some kind of tolerance https://github.com/adamantine-sim/adamantine/blob/1a42819437d702713b351cddbeb1287e2b232dfe/source/experimental_data.cc#L592-L608
 2. The tolerance used here https://github.com/adamantine-sim/adamantine/blob/1a42819437d702713b351cddbeb1287e2b232dfe/source/experimental_data.cc#L567-L570
 should be a relative tolerance not an absolute one.

Rombur:
Let's close this issue for now. This code has changed significantly. We can reopen it if the problem reappears.

PR #229, opened 2023-03-13
Refactor Raytracing

Rombur:
@stvdwtt can you try this PR and tell me if it fixes the problem in #225

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/374/pipeline)

### DIFF
--- a/source/RayTracing.cc
+++ b/source/RayTracing.cc
@@ -6,6 +6,7 @@
  */
 
 #include <RayTracing.hh>
+#include <utils.hh>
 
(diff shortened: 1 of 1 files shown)


Rombur:
Can you try now. It should fix the problem. I did a force push on the branch sorry :(

stvdwtt (approved):
There still are some details that we might have to sort out for elements along the edge, but that can/should wait until we have better visualization of the experimental data.

PR #234, opened 2023-06-16
Remove dealii-weak_forms from the infrastructure

Rombur:
This PR removes dealii-weak_forms. Everything seems to be working. The new code is very close to the old one. There is a `FIXME` that I will probably need help with.

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/386/pipeline)


Rombur:
@stvdwtt we can discuss the FIXME later (tomorrow?). The current code does the same thing as the old one so I think this can be merged as is.

stvdwtt:
@Rombur: Oh, I misunderstood the comment then. So the code in the PR reproduces the old code behavior, but doesn't match an alternative formulation that you expect to give the same result? Talking tomorrow about it is a good idea. I'll merge this in.

Rombur:
>  So the code in the PR reproduces the old code behavior, but doesn't match an alternative formulation that you expect to give the same result?

yes, exactly. You can manipulate the equation and get a simpler form but when I code it, I get a different result.

PR #236, opened 2023-07-07
Experiment fixes/improvements

stvdwtt:
This PR includes an assortment of small changes and fixes to reading experimental data and doing data assimilation.

Summary:
- Change definition of experimental "rays" to match ROS definition
- Add the option to output the experimental data projected onto the mesh
- Add an option to refine the mesh for the output (useful for a more accurate representation of fields with higher order elements)
[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/387/pipeline)


application/adamantine.hh
```diff
@@ -1789,11 +1797,37 @@ run_ensemble(MPI_Comm const &communicator,
             thermal_physics_ensemble[0]->get_dof_handler();
         auto expt_to_dof_mapping = adamantine::get_expt_to_dof_mapping(
             points_values, thermal_dof_handler);
+        std::cout << "Number expt sites mapped to DOFs: "
+                  << expt_to_dof_mapping.first.size() << std::endl;
```
Rombur, line 1801: If you want this, you should put it inside `if(rank==0)`.

application/adamantine.hh
```diff
@@ -1789,11 +1797,37 @@ run_ensemble(MPI_Comm const &communicator,
...
+        bool const output_experiment_on_mesh =
+            experiment_optional_database.get().get("output_experiment_on_mesh",
+                                                   true);
+        if (output_experiment_on_mesh)
+        {
+          dealii::LA::distributed::Vector<double, MemorySpaceType>
+              temperature_expt;
+          temperature_expt.reinit(
+              solution_augmented_ensemble[0].block(base_state));
+          temperature_expt.add(1.0e10);
```
Rombur, line 1818: I slightly prefer using a negative temperature but that's fine too :-)

application/adamantine.hh
```diff
@@ -1492,6 +1492,14 @@ run_ensemble(MPI_Comm const &communicator,
             thermal_physics_ensemble[member]->get_dof_handler(), member));
   }
 
+  // PostProcessor for outputting the experimental data
+  boost::property_tree::ptree post_processor_expt_database;
+  post_processor_expt_database.put("filename_prefix", "expt");
```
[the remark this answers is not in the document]
stvdwtt: For context, here's how the current way works:
"Regular" output files: [output_prefix]_[ensemble_index].[mpi_task].[timestep_number].vtu
Experiment output files: [expt_output_prefix].[mpi_task].[timestep_number].vtu

That is, the experiment prefix is entirely independent of the prefix for regular outputs. Currently the experiment prefix is hard coded in `adamantine.hh`.

Do you have a preference between:
1) Keep the regular output prefix and experiment output prefix independent, but expose the experiment output prefix to the user so they can choose the same starting characters if they like (e.g. "output" and "output_expt").
2) Restructure how the experiment output by appending the experiment string after the output prefix (i.e. [output_prefix].[expt_output_prefix].[mpi_task].[timestep_number].vtu).
3) Restructure how the experiment output by appending the experiment string at the very end (i.e. [output_prefix].[mpi_task].[timestep_number].[expt_output_prefix].vtu).

VisIt can properly group all of these options, so I'm ok with any of them.

application/adamantine.hh
stvdwtt: Agreed, I'll make this change.

application/adamantine.hh
stvdwtt: I started with a negative temperature, but switched to this so that I could independently threshold T < 1 as FE_Nothing cells and T > 100000 as non-FE_Nothing cells with no data.

I could switch it to a negative number and then do that first threshold as -1 < T < 1, which probably is worth doing.

source/experimental_data_utils.cc
```diff
@@ -104,6 +104,12 @@ void set_with_experimental_data(
     std::pair<std::vector<int>, std::vector<int>> &expt_to_dof_mapping,
     dealii::LinearAlgebra::distributed::Vector<double> &temperature)
 {
+  std::cout << "Number of unique experimental points is "
+            << std::set<double>(expt_to_dof_mapping.second.begin(),
+                                expt_to_dof_mapping.second.end())
+                   .size()
+            << std::endl;
```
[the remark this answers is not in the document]
stvdwtt: Sounds good, I'll make that change.

application/adamantine.hh
Rombur: ah I see that makes sense. Keep as it is then

stvdwtt:
Updated CI:
https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/389/pipeline

source/experimental_data_utils.cc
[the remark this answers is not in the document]
stvdwtt: The verbose parameter was specifically for refinement, so I changed it to a global verbosity flag and hid these lines behind it.

Rombur (approved):

PR #237, opened 2023-08-01
Wait for files to appear instead of throwing an error

Rombur:
This PR does two things:
 1. replace `boost::filesystem` with `std::filesystem`. `std::filesystem` is part of C++17 but before gcc 9 it was implemented in a separate library that we had to link against. Since gcc 8 is quite old (Ubuntu 20.04 is using gcc 9.3), it's safe to make the chang.
 2. wait for the input files to appear instead of throwing an error

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/388/pipeline)


stvdwtt (approved):
Great! I'm glad the interface for waiting for all those files is unified and that it prints out the name of the file it is waiting for. That had been a pain point earlier if I had a typo somewhere in a filename.

PR #238, opened 2023-08-10
Add lumps of material during deposition

Rombur:
With this PR we can add more materials at once. I choose a deposition time parameter but we can pick something else. I don't know what's easier to use. One thing we need to be careful about is that our source term currently does not support multiple layers at the same time. So if we add too much material at the end of the layer the source term will be wrong.

[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/390/pipeline)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ The following options are available:
         * deposition\_width: width of material deposition boxes (in the plane of the material, normal to the scan direction, 3D only)
         * deposition\_height: height of material deposition boxes (out of the plane of the material)
         * deposition\_lead\_time: amount of time before the scan path reaches a point that the material is added
+        * deposition\_time: using this option, the material is added in bigger lumps (optional)
   * import\_mesh: true or false (required)
   * if import\_mesh is true:
     * mesh\_file: The filename for the mesh file (required)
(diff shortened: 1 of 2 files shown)


Rombur:
DO NOT MERGE: I found a bug.

Rombur:
The CI just started to work again by itself: [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/392/pipeline)

stvdwtt (commented):
@Rombur Apologies for the late review.

Although the CI passes, if I try to pull this branch and run the SNS Build2 case as-is, I get an error saying that it can't find the mesh. I believe this was fixed by #240.

If I merge master into this branch, that error is gone (as expected). However, the lack of mesh activation bug fixed by #248 crops up again. Since I think you're touching similar code here as there, I wonder if the automatic merge isn't getting it quite right.

That all said, it tested the lumped material addition and it works as expected, which is fantastic.

If you can merge the changes from #248 into this and retain the expected behavior, then I'll approve and merge this in.

Rombur:
Can you try again? I just rebased the branch, it seems to be working fine. When you merged with `master`, was your `master` up-to-date?

stvdwtt:
My master was up to date, but this updated branch is working for me. Maybe your rebase did something different than when I merged?

In any case, this now works for me, so I'm merging it in.

PR #241, opened 2023-08-15
Fix risk of overflow in PostProcessor

Rombur:
[CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/397/pipeline)

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -56,8 +56,8 @@ template <int dim, typename MemorySpaceType,
               std::is_same<MemorySpaceType, dealii::MemorySpace::Host>::value,
               int> = 0>
 void output_pvtu(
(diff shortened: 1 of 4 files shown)


stvdwtt:
This is perhaps out of scope for this PR, but do you know what the intent for the "cycle" input to the filename is? I'm not seeing it incremented in the app at all, so I think it is always zero.

Rombur:
Oh let me remove it. It's supposed to be the refinement cycle but we don't refine at a fixed time, so it's useless.

Rombur:
New [CI](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/398/pipeline)

stvdwtt (approved):

PR #243, opened 2023-09-01
Add support for J2 plasticity 

Rombur:
This is still a work in progress. The first two commits could be moved to their own PRs but they are literally one line each so I don't think it matters.
Things left to do:
- Decide where the new code should go. I've put it in `MechanicalPhysics` but we could put it in `MechanicalOperator`. My reasoning is that, the new code does not need "solving" but I don't know where is the right place.
- The code does not work if the mesh is refined or material is added.
- Testing. I tried to find analytical solutions on internet but I couldn't find anything. Usually people just do comparisons with commercial software but the models are slightly different and the results don't match. If I could get help with that, it would be fantastic.

### DIFF
--- a/source/MechanicalPhysics.cc
+++ b/source/MechanicalPhysics.cc
@@ -301,9 +301,8 @@ void MechanicalPhysics<dim, MemorySpaceType>::compute_stress(
        _dof_handler.active_cell_iterators() |
            dealii::IteratorFilters::ActiveFEIndexEqualTo(0, true))
   {
-    // Formulation based on the combines isotropic-kinematic hardening model for
-    // J2 plasticity on the combined isotropic-kinematic hardening model for J2
-    // plasticity in Chapter 3 of R. Borja, Plasticity: Modeling and
+    // Formulation based on the combined isotropic-kinematic hardening model for
+    // J2 plasticity in Chapter 3 of R. Borja, Plasticity: Modeling and
     // Computation, Springer-Verlag, 2013. DOI: 10.1007/978-3-642-38547-6
     //
     // Compute the strain. We get the strain for all the quadrature points at
--- a/source/MechanicalPhysics.cc
+++ b/source/MechanicalPhysics.cc
@@ -8,9 +8,13 @@
 #include <MechanicalPhysics.hh>
 #include <instantiation.hh>
 
+#include <deal.II/base/symmetric_tensor.h>
+#include <deal.II/base/tensor.h>
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/fe/fe_nothing.h>
 #include <deal.II/fe/fe_q.h>
+#include <deal.II/hp/fe_values.h>
+#include <deal.II/lac/la_parallel_vector.h>
 #include <deal.II/lac/solver_cg.h>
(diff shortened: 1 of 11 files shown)


stvdwtt:
Great!

Re where it goes: I'll read it through let you know if I have any thoughts/preferences.

Re testing: This is mostly a note for myself. Verification Test 1 mentioned here might be of use, but I need to read through it in more detail.
https://csm.mech.utah.edu/content/wp-content/uploads/2011/10/9tutorialOnJ2Plasticity.pdf

Rombur:
Analytical solution on sphere and cyclinder:
- https://doi.org/10.1016/0308-0161(94)90096-5
- https://doi.org/10.1016/0308-0161(93)90004-D

Rombur:
This PR should actually work if you add material in serial. I need to check that it works properly though.

Rombur:
I think this is PR should be good enough for what Ayana needs (ie we can add material in serial but not use AMR). Of course, we still don't have good tests so that's pretty annoying.

stvdwtt:
@Rombur, sounds good. I'll review and merge as soon as possible. The test case for Ayana will at least give us a sanity check until we have good tests in place.

source/MechanicalPhysics.cc
```diff
@@ -94,49 +116,257 @@ void MechanicalPhysics<dim, MemorySpaceType>::setup_dofs(
...
+  const dealii::FEValuesExtractors::Vector displacement_extr(0);
+  unsigned int cell_id = 0;
+  for (auto const &cell :
+       _dof_handler.active_cell_iterators() |
+           dealii::IteratorFilters::ActiveFEIndexEqualTo(0, true))
+  {
+    // Formulation based on the combines isotropic-kinematic hardening model for
+    // J2 plasticity on the combined isotropic-kinematic hardening model for J2
+    // plasticity in Chapter 3 of R. Borja, Plasticity: Modeling and
+    // Computation, Springer-Verlag, 2013. DOI: 10.1007/978-3-642-38547-6
```
stvdwtt, line 307:
```suggestion
    // Formulation based on the combined isotropic-kinematic hardening model for
    // J2 plasticity in Chapter 3 of R. Borja, Plasticity: Modeling and
    // Computation, Springer-Verlag, 2013. DOI: 10.1007/978-3-642-38547-6
```

stvdwtt:
@Rombur There was a repeated phrase in one of the comments (I think probably due to how I structured the "suggested edit"). I went ahead and fixed that directly. Once the CI passes I'll merge this in.

Rombur:
There was a failure (see [here](https://cloud.cees.ornl.gov/jenkins-ci/job/adamantine/job/PR-243/4/testReport/junit/projectroot/tests/Test___CPU_on_device___test_thermal_operator_1/)). I have seen this a couple of times. Usually re-running fixes the issue. I don't know where the problem comes from. I have run the test in valgrind and everything looks fine.

Rombur:
The CI passed, so we can merge this. 

@stvdwtt where do you think the new code should go? MechanicalPhysics, MechanicalOperator, or a new Operator? I can move the code in a different PR if we think it would be better somewhere else.

PR #245, opened 2023-09-19
fix file parsing logic in load_segment_scan_path

Lance-Drane:
Hi, I'm part of the INTERSECT SDK team - I help a little bit with integrating the adamantine digital twin into the overall INTERSECT deployment. [Marshall McDonnell](https://github.com/marshallmcdonnell/) and Jesse McGaha can vouch for me. While debugging the deployment, Jesse and I noticed a few oddities with the segment file scanner:

- The error output message is 0-indexed, but human-readable file lines are generally expected to be 1-indexed. i.e. I was getting an error output complaining about line 17 when it really meant line 18. The data_index variable remains 0-indexed, but shifts to being 1-indexed when the error is thrown.
- When scanning the path segment lines, it makes more sense to use `&&` instead of `||`. As currently implemented, the entire file will always be scanned, and the value of `n_segments` is never relevant. My change fixes this.
- in the current implementation, if `&&` were implemented correctly, `n_segments` was also subject to the off-by-one error. For example, if `number of path segments` was evaluated to `1`, none of the path segments would have been executed.


.gitignore
```diff
@@ -0,0 +1 @@
+.vscode
```
Rombur, line 1: Please don't add a `.gitignore`

Rombur (approved):
Thanks. Looks good, just don't add the `.gitignore` file.

Rombur:
OK to test

Rombur:
We now have a test failing (see [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/406/tests)). I think this because, as you pointed out, the number of segments was irrelevant before this PR. Can you change the number of segments from 2 to 3 in [tests/data/scan_path_L.txt](https://github.com/adamantine-sim/adamantine/blob/a57da74709d7be4cbcdedbc7cfebb86da51aff8d/tests/data/scan_path_L.txt#L2C1-L2C2)

Lance-Drane:
> We now have a test failing (see [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/adamantine/detail/adamantine/406/tests)). I think this because, as you pointed out, the number of segments was irrelevant before this PR. Can you change the number of segments from 2 to 3 in [tests/data/scan_path_L.txt](https://github.com/adamantine-sim/adamantine/blob/a57da74709d7be4cbcdedbc7cfebb86da51aff8d/tests/data/scan_path_L.txt#L2C1-L2C2)

Happy to make this change. There's another test file, `integration_da_add_material_sp.txt`, which will now only analyze the first 3 data lines instead of all of them. Would you like for me to change the number of path segments here to `13` instead of the initial value of `3`?

There are a couple of other files `tests/data/thermoelastic_bare_plate_add_material_scan_path.txt` and `tests/data/thermoelastic_bare_plate_scan_path.txt` which say that the number of path segments is `3` but there are only 2 lines of data; however, this shouldn't affect any tests, as the whole file will continue to be parsed. Should those files be updated, or are they fine as-is?

Rombur:
Yes, please update all the files.

Rombur:
Thanks for the fix!

PR #251, opened 2023-12-05
Fix parsing of the elastic limit

stvdwtt:
MaterialProperty was checking each state for `elastic_limit`, and setting it to infinity if it wasn't there. The result was that the elastic limit for the solid was ignored.

This changes the logic so MaterialProperty only sets the elastic limit to infinity if it isn't listed for the solid.


stvdwtt:
Retest this please.

Rombur (approved):

PR #253, opened 2024-01-30
Update dependencies and use Kokkos

Rombur:
This PR updates the dependencies that we are using. In particular, we bump the required version of deal.II to 9.5. This version of deal.II replaced the CUDA backend with Kokkos, so I removed our CUDA code and replaced it with Kokkos. There are probably more places where the code could be simplified but that can wait.


Rombur:
Retest this please

stvdwtt (approved):

PR #254, opened 2024-02-01
Changed eps in timestepping 

AshGannon:
We use an epsilon to get the "expected" behavior when the deposition time and the time match should match exactly but don't because of floating point accuracy. I changed eps to be computed using the deposition start time and error between the time step and activation time.

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -649,7 +649,6 @@ void refine_mesh(
   // PropertyTreeInput refinement.n_heat_refinements
   unsigned int const n_kelly_refinements =
       refinement_database.get("n_heat_refinements", 2);
-  std::cout << "Kelly refinements the first time" << n_kelly_refinements << std::endl;
   double coarsening_fraction = 0.3;
   double refining_fraction = 0.6;
   // PropertyTreeInput refinement.heat_cell_ratio
@@ -1021,15 +1020,6 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   double const duration = time_stepping_database.get<double>("duration");
 
   // Extract the refinement database
-  // Refinement database contains the following entries:
-  // - n_heat_refinements: number of times the mesh will be refined using the
-  //   Kelly error estimator
-  // - n_beam_refinements: number of times the mesh will be refined along the
-  //   beam paths
-  // - heat_cell_ratio: fraction of cells that will be refined using the Kelly
-  //   error estimator
-  // - max_level: maximum level of refinement
-  // - beam_cutoff: cut-off value for the heat source. Cells where the heat
   boost::property_tree::ptree refinement_database =
       database.get_child("refinement");
   // PropertyTreeInput refinement.time_steps_between_refinement
@@ -1042,8 +1032,6 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
   // PropertyTreeInput materials.new_material_temperature
   double const new_material_temperature =
       database.get("materials.new_material_temperature", 300.);
-  unsigned int const n_kelly_refinements =
-      refinement_database.get("n_heat_refinements", 2);
 
 #ifdef ADAMANTINE_WITH_CALIPER
   CALI_CXX_MARK_LOOP_BEGIN(main_loop_id, "main_loop");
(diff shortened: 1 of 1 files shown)


application/adamantine.hh
```diff
@@ -649,6 +649,7 @@ void refine_mesh(
   // PropertyTreeInput refinement.n_heat_refinements
   unsigned int const n_kelly_refinements =
       refinement_database.get("n_heat_refinements", 2);
+  std::cout << "Kelly refinements the first time" << n_kelly_refinements << std::endl;
```
Rombur, line 652: If you want to add a screen output, you should guard it with `if(rank ==0)` otherwise every processor will write on the screen. Currently if you run with 10 processors, you will get the message 10 times.

Note that you probably don't want to use the Kelly estimator-based refinement strategy. This was the first strategy I used but it doesn't work very well in our cases. I  want to remove this option in the future.

application/adamantine.hh
```diff
@@ -1093,8 +1111,7 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
           // activation_end.
           timers[adamantine::add_material_search].start();
           auto elements_to_activate = adamantine::get_elements_to_activate(
-              thermal_physics->get_dof_handler(), material_deposition_boxes);
-
```
Rombur, line 1097: Unless you had a reason, undo the blank line change here and below

application/adamantine.hh
```diff
@@ -1020,6 +1021,15 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
...
   // Extract the refinement database
+  // Refinement database contains the following entries:
+  // - n_heat_refinements: number of times the mesh will be refined using the
+  //   Kelly error estimator
+  // - n_beam_refinements: number of times the mesh will be refined along the
+  //   beam paths
+  // - heat_cell_ratio: fraction of cells that will be refined using the Kelly
+  //   error estimator
+  // - max_level: maximum level of refinement
+  // - beam_cutoff: cut-off value for the heat source. Cells where the heat
```
Rombur, line 1032: I understand why you want this in the code but the information is already in the README and on the website. If we add it here, we now have three places to update.

application/adamantine.hh
```diff
@@ -1051,7 +1062,7 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
     // is necessary when using an embedded method.
     if ((((n_time_step % time_steps_refinement) == 0) ||
          (time >= next_refinement_time)) &&
-        use_thermal_physics)
+        use_thermal_physics && n_kelly_refinements > 0)
```
Rombur, line 1065: This seems wrong. Why do you want to refine the mesh only when using the Kelly refinement?

AshGannon:
I only ended up changing the denominator of eps.

Rombur (approved):

stvdwtt:
@AshGannon do you have a sense for when the epsilon is too small and when it isn't? Are we safe hard-coding it like you have, or should it depend on something problem specific?

Rombur:
The tolerance already depends on the problem through the time step. It's obviously not ideal to have a "magic" number in the code but when dealing with round-off you usually don't have a choice.

The PR is not passing the CI but that's because we need to merge #253 first.

AshGannon:
> @AshGannon do you have a sense for when the epsilon is too small and when it isn't? Are we safe hard-coding it like you have, or should it depend on something problem specific?

I would like for it to depend on something problem specific, but I haven't quite figured out a way to make to do so without hard-coding anything. I tried setting it to the difference between dt and the new time step - the old time step. This was too small, and the grid didn't properly activate in the hourglass simulation - even in the beginning. 

Bruno mentioned to me that he originally chose 1e12 because you can expect an accuracy of about 12 digits with double precision. He also uses this eps for power and time in deposition_along_scan_path. In my hourglass simulation, the dwell time is set to 0, so my deposition_time.begin() happened to be 1e-12. The deposition_time was also not set. Since double eps = (*deposition_times.begin())/abs(activation_time_end - time); worked in this example, double eps = 1e-12/(time - time_old) might work generally. It's the same as dt/1e12, but you accumulate error in dt as time goes on.

stvdwtt:
Retest this please

AshGannon:
@stvdwtt I'm not sure what you mean. Retest the 1e-10? Or the dt/time-time_old? I ran the latter over the weekend, the mesh fully activated and the results were very close at the times I checked (just spot checked). Are you seeing something odd with the 1e-10?

stvdwtt:
@AshGannon Sorry! That wasn't directed at you -- that's the command for the continuous integration tests to run again. I just merged in Bruno's PR and my understanding is that the tests in this PR should pass now. Assuming they do, we'll merge this.

PR #255, opened 2024-02-07
Add json reader

Rombur:
With this PR, we can read input files that use the `json` format. It's easier to manipulate `json` file than `info` file in python so it should make your life easier @stvdwtt


stvdwtt (approved):
Thanks for doing this! It also will make Mist support for material properties much easier.

PR #261, opened 2024-02-28
Reduce max FE degree from 10 to 5

Rombur:
I am going to introduce a few more template parameters, so let's reduce the number of instantiations before I do that. This PR also adds the `-ffast-math` flag when compiling the `adamantine` binary in the Docker image.


stvdwtt (approved):
Looks good, I'll merge once the CI is done.

Rombur:
I forgot to update `validate_input_database`. I just pushed to fix it.

Rombur:
The CUDA build is not passing but it's due to issue with the CI machines. There is no ETA on when this will be fixed.

stvdwtt:
Merging without the CUDA tests running since we don't know when the CI will be fixed. @Rombur manually ran the tests for HIP and they passed.

PR #262, opened 2024-03-01
Make MaterialProperty::polynomial_order a template parameter

Rombur:
Currently when using a polynomial description of the material properties, we assume that we have a polynomial of order four. This is often not the case and we end up wasting a lot of flops. This PR makes the polynomial order a template parameter. The correct order is computed at runtime by doing a first read of the material parameters before calling `MaterialProperties`. The new template parameter `p_order` propagates everywhere because we need to template `MaterialProperties` and almost every class owns a reference to `MaterialProperties`.

### DIFF
--- a/application/adamantine.cc
+++ b/application/adamantine.cc
@@ -25,6 +25,83 @@
 #include <caliper/cali-manager.h>
 #endif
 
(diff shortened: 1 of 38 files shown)


Rombur:
Retest this please

PR #266, opened 2024-03-15
Increase flexibility in creating meshes and loading meshes

stvdwtt:
This is the first of two PRs to let us do the curved wall build for the Okuma. This one is focused on relaxing the assumption that the entire domain is in the +x, +y, +z octant. A common AM reference frame has the top of the substrate centered at (0, 0, 0).

Changes:
- Removes the check that the material height must be positive
- Allows the user to specify the origin for an adamantine-generated meshes
- Allows user-defined scaling of a read-in mesh
- Allows the material id of a read-in mesh be cleared and set to 0

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,14 +106,14 @@ The following options are available:
     vtu, exodus, or default, i.e., use the file suffix to try to determine the
     mesh format (required)
     * mesh\_scale\_factor: Apply a uniform scaling factor to the mesh (e.g. if the mesh is defined in mm or inches instead of m), (default value: 1)
-    * set\_material\_id\_to\_0: Clear the material IDs defined in the mesh and sets them all to zero: true or false (default value: false)
+    * reset\_material\_id: Clear the material IDs defined in the mesh and set them all to zero so all material properties are given by the `material\_0` input block: true or false (default value: false)
   * if import\_mesh is false:
     * length: the length of the domain in meters (required)
     * height: the height of the domain in meters (required)
     * width: the width of the domain in meters (only in 3D)
-    * length_origin: the reference location in the length direction (default value: 0)
-    * height_origin: the reference location in the height direction (default value: 0)
-    * width_origin: the reference location in the width direction (only in 3D) (default value: 0)
+    * length\_origin: the reference location in the length direction (default value: 0)
+    * height\_origin: the reference location in the height direction (default value: 0)
+    * width\_origin: the reference location in the width direction (only in 3D) (default value: 0)
     * length\_divisions: number of cell layers in length (default value: 10)
     * height\_divisions: number of cell layers in the height (default value: 10)
     * width\_divisions: number of cell layers in width (only in 3D) (default value: 10)
--- a/README.md
+++ b/README.md
@@ -105,10 +105,15 @@ The following options are available:
     * mesh\_format: abaqus, assimp, unv, ucd, dbmesh, gmsh, tecplot, xda, vtk,
     vtu, exodus, or default, i.e., use the file suffix to try to determine the
     mesh format (required)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -105,10 +105,15 @@ The following options are available:
     * mesh\_format: abaqus, assimp, unv, ucd, dbmesh, gmsh, tecplot, xda, vtk,
     vtu, exodus, or default, i.e., use the file suffix to try to determine the
     mesh format (required)
+    * mesh\_scale\_factor: Apply a uniform scaling factor to the mesh (e.g. if the mesh is defined in mm or inches instead of m), (default value: 1)
+    * set\_material\_id\_to\_0: Clear the material IDs defined in the mesh and sets them all to zero: true or false (default value: false)
   * if import\_mesh is false:
     * length: the length of the domain in meters (required)
     * height: the height of the domain in meters (required)
     * width: the width of the domain in meters (only in 3D)
+    * length_origin: the reference location in the length direction (default value: 0)
```
Rombur, line 114:
```suggestion
    * length\_origin: the reference location in the length direction (default value: 0)
```

README.md
```diff
@@ -105,10 +105,15 @@ The following options are available:
...
+    * height_origin: the reference location in the height direction (default value: 0)
```
Rombur, line 115:
```suggestion
    * height\_origin: the reference location in the height direction (default value: 0)
```

README.md
```diff
@@ -105,10 +105,15 @@ The following options are available:
...
+    * width_origin: the reference location in the width direction (only in 3D) (default value: 0)
```
Rombur, line 116:
```suggestion
    * width\_origin: the reference location in the width direction (only in 3D) (default value: 0)
```

README.md
```diff
@@ -105,10 +105,15 @@ The following options are available:
     * mesh\_format: abaqus, assimp, unv, ucd, dbmesh, gmsh, tecplot, xda, vtk,
     vtu, exodus, or default, i.e., use the file suffix to try to determine the
     mesh format (required)
+    * mesh\_scale\_factor: Apply a uniform scaling factor to the mesh (e.g. if the mesh is defined in mm or inches instead of m), (default value: 1)
+    * set\_material\_id\_to\_0: Clear the material IDs defined in the mesh and sets them all to zero: true or false (default value: false)
```
Rombur, line 109: I think I would call this something like `set_material_id` or `set_uniform_material_id`. Use don't need to know/don't care which value we are using.

Rombur (commented):
The PR looks fine. I am not sure about the name `set_material_id_to_0`

README.md
stvdwtt: How about `reset_material_id`, and then mention that we set it to zero in the description? I think the value being zero is important because it corresponds to the `material_0` material input block.

```suggestion
    * reset\_material\_id: Clear the material IDs defined in the mesh and set them all to zero so all material properties are given by the `material\_0` input block: true or false (default value: false)
```

Rombur: Sounds good.

source/Geometry.cc
```diff
@@ -88,6 +89,22 @@ Geometry<dim>::Geometry(MPI_Comm const &communicator,
...
     _triangulation.copy_triangulation(serial_triangulation);
+
+    // Apply user-specified scaling to the mesh
+    // PropertyTreeInput geometry.mesh_scale_factor
+    auto mesh_scaling = database.get("mesh_scale_factor", 1.0);
+    dealii::GridTools::scale(mesh_scaling, _triangulation);
+
+    // Set the mesh material id to 0 if specified in the input
+    // PropertyTreeInput geometry.set_material_id_to_0
+    auto reset_material_id = database.get("set_material_id_to_0", false);
```
Rombur, line 100: This still needs to be updated

Rombur (approved):

PR #267, opened 2024-03-19
Change the scan-path-based deposition convention

stvdwtt:
This PR aligns the scan-path-based material deposition convention with Abaqus and is generally more intuitive.

Originally, the convention was the the scan path goes along the "old" height of the material and additional material is added on top of it. This is a bit counter-intuitive because the actual center of the heat source is at the top of the added material. (This is controlled by the `height` variable that is passed into `HeatSource::value`.) This convention also goes against the convention that Abaqus uses, where the z-height of the event series in the scan path is the top of the heat source.

This PR shifts the convention so that the material deposition box starts at the z-height given in the scan path and then adds material downward. That change is only two lines. Most of this PR is updating tests and adding some descriptive text in the README.


Rombur:
Retest this please

Rombur (approved):
Looks good.

PR #269, opened 2024-03-26
Add support to using more than one MPI rank per ensemble member

Rombur:
This PR enables us to use multiple processors per ensemble member. This only works if the all the ensemble members use the same number of processors. This ensures that the mesh is partitioned the same way for all the ensemble members which simplifies some operations. We could remove this condition in the future.

The changes in the code are pretty small if you hide the whitespace changes. Most of the bugs were related to some data not being distributed to all the processors.

### DIFF
--- a/ci/jenkins_config
+++ b/ci/jenkins_config
@@ -22,7 +22,7 @@ pipeline {
             docker {
               image "rombur/adamantine-stack:no_gpu-latest"
                 alwaysPull true
-                label 'docker || nvidia-docker || rocm-docker'
+                label 'nvidia-docker || rocm-docker'
             }
           }
           steps {
@@ -38,7 +38,7 @@ pipeline {
                   -D DEAL_II_DIR=${DEAL_II_DIR} \
                   ..
                 '''
-                sh 'make -j3'
+                sh 'make -j8'
                 sh 'ctest --no-compress-output -R test_ -T Test'
               }
           }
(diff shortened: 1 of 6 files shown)


Rombur:
This requires #268. Otherwise we get too many warnings and the compilation stops

stvdwtt:
@Rombur, got it, I just merged in #268.

Rombur:
Retest this please

Rombur:
Retest this please

ci/jenkins_config
```diff
@@ -38,7 +38,7 @@ pipeline {
                   -D DEAL_II_DIR=${DEAL_II_DIR} \
                   ..
                 '''
-                sh 'make -j3'
+                sh 'make -j2'
```
Rombur, line 41: We need a little bit less than 3GB per process but the CPU machine only have 8GB of RAM. Reducing the number of processors fixed the compilation error.

ci/jenkins_config
```diff
@@ -22,7 +22,7 @@ pipeline {
             docker {
               image "rombur/adamantine-stack:no_gpu-latest"
                 alwaysPull true
-                label 'docker || nvidia-docker || rocm-docker'
+                label 'nvidia-docker || rocm-docker'
```
Rombur, line 25: The CPU nodes don't have enough memory and cores. Instead use a GPU node

stvdwtt (approved):

PR #270, opened 2024-04-11
Remove embedded time stepping methods and Kelly estimator AMR

Rombur:
This is the last PR I would like to get merged before creating the release branch. This removes the embedded time stepping methods (aka the adaptive time step method) and the AMR based on the Kelly estimator. We are not using either of these capabilities and they are making the code and the input file more complicated. The embedded method methods are problematic for the GPU computations and the Kelly estimator AMR is not usable in practice because the number of cells cannot be kept stable. We may reintroduce these capabilities in the future but we need to do it in a more thoughtful way.

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -973,8 +973,10 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
     if ((time + time_step) > duration)
       time_step = duration - time;
 
-    // Refine the mesh after time_steps_refinement time steps.
-    if (((n_time_step % time_steps_refinement) == 0) && use_thermal_physics)
+    // Refine the mesh the first time we get in the loop and after
+    // time_steps_refinement time steps.
+    if (((n_time_step == 1) || ((n_time_step % time_steps_refinement) == 0)) &&
+        use_thermal_physics)
     {
       timers[adamantine::refine].start();
       double next_refinement_time = time + time_steps_refinement * time_step;
@@ -983,8 +985,11 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
                   time_steps_refinement, refinement_database);
       timers[adamantine::refine].stop();
       if ((rank == 0) && (verbose_output == true))
-        std::cout << "n_dofs: " << thermal_physics->get_dof_handler().n_dofs()
+      {
+        std::cout << "n_time_step: " << n_time_step << " time: " << time
+                  << " n_dofs: " << thermal_physics->get_dof_handler().n_dofs()
                   << std::endl;
+      }
     }
 
     // Add material if necessary.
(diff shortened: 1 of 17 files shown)


Rombur:
@stvdwtt _ping_

application/adamantine.hh
```diff
@@ -717,15 +634,6 @@ void refine_mesh(
...
     }
 
-    // Flag the cells for refinement.
-    for (auto &cell : cells_to_refine)
-    {
-      if (coarsen_after_beam)
-        cell->clear_coarsen_flag();
-      if (cell->level() < max_level)
-        cell->set_refine_flag();
-    }
```
stvdwtt, line 727: Previously we had this loop that cleared the coarsen flag for the cells that had the refine flag set. It looks like with this change all cells will have the coarsen flag set and some will additionally have the refine flag set. I couldn't easily find in the docs what happens if both are set for one cell.

I get that we don't need to keep the level check, but I wonder if we still need 

```// Flag the cells for refinement.
    for (auto &cell : cells_to_refine)
    {
      if (coarsen_after_beam)
        cell->clear_coarsen_flag();
    }

Comment (Rombur) [reply application/adamantine.hh:727]: Good catch. This is actually totally wrong. We don't refine the cells anymore because we don't use `cells_to_refine` anywhere. The testsuite needs improvements...

File: application/adamantine.hh [hunk 2]
```diff
@@ -722,8 +639,8 @@ void refine_mesh(
     {
       if (coarsen_after_beam)
         cell->clear_coarsen_flag();
-      if (cell->level() < max_level)
-        cell->set_refine_flag();
+
+      cell->set_refine_flag();
```
Rombur, line 643: Here is the fix.

Rombur:
Retest this please

application/adamantine.hh
stvdwtt: I'm going to test this manually and then add that test in another PR. (Unless you already did a new manual test case for it.)

Rombur: Please do

stvdwtt: With the max refinement check removed, each time AMR is called cells near the beam get refined again, even if they are already refined. We should add the check back, but based on `n_refinements` instead of `max_level`.

(This assumes that we want to merge the concept of how many times we refine each time we do AMR and the max level. In principle one could want to keep them seperate and slowly get to the max refinement over a few AMR cycles, but that wouldn't make much sense in our context.)

Rombur: Thanks. I fixed it.

stvdwtt: It now refines to the expected level and coarsens, but for the specific test case I tried with added material I'm different amounts of added material. I need to look further into it, but I'm wondering if the order of refinement and deposition were swapped.

stvdwtt: Here are the input files and scan path file for the test cases. I had to add a `.txt` extension on the input files for GitHub to let me upload them.

[master_input.info.txt](https://github.com/adamantine-sim/adamantine/files/15044845/master_input.info.txt)
[demo_316_short_scan_path.txt](https://github.com/adamantine-sim/adamantine/files/15044846/demo_316_short_scan_path.txt)
[pr_input.info.txt](https://github.com/adamantine-sim/adamantine/files/15044847/pr_input.info.txt)

Rombur: The last commit should fix the issue. I have checked and the number of dofs is now the same for this PR and master.

stvdwtt (approved):
I re-ran the manual test and the result looks good. I had to modify the test case to avoid activating parts of coarse cells.

Issue #274, opened 2024-04-22
Disallow coarsening if any children of the parent cell are FE_Nothing


stvdwtt:
Context: With AMR activated we can add material in refined areas that corresponds to some but not all of the child elements of a element in the coarse mesh.

Current behavior: As the heat source moves away from the area, the code tries to coarsen the mesh in the deposited region and ends up with NANs as it tries to merge active and inactive elements.

Desired behavior: The mesh stays refined if any other children of the parent cell are FE_Nothing.

PR #276, opened 2024-04-23
Add AMR integration test with coarsening enabled

stvdwtt:
While debugging #270 we realized that we don't have a test where the mesh is allowed to coarsen after the beam. This PR adds a test for that. 

Note: Due to the issue with mesh coarsening across active/inactive boundaries (#274) this test case had to be carefully chosen. The fix to #274 should have a similar test to this one but with a shifted scan path to force active/inactive boundaries across children of the same coarse mesh element.


Rombur (approved):

Rombur:
The CI error is unrelated to this PR. I did not know this but `fetnat08` and `fetnat09` only have four cores while all the others GPU machines have at least six. The error in the CI is from MPI because we are trying toe oversubscribe the machine.

PR #279, opened 2024-04-30
Print all deal.II configurations problems before aborting

masterleinad:
I needed to reconfigure deal.II multiple times since I didn't look into the file that checks the configuration. It's easier to print all the configuration problems at once.


Rombur:
If you want to print all the configuration issues at once, then accumulate the options that are not satisfied in one string and print one error message. With this PR, we are printing the same sentence many times over, only is a single word is changed. Moreover the last sentence does not give any new information, it just fills the screen.

cmake/SetupDealII.cmake
```diff
@@ -3,26 +3,20 @@ find_package(deal.II 9.5 REQUIRED PATHS ${DEAL_II_DIR})
...
-endif()
+set(DEAL_II_REQUIRED_FEATURES ARBORX CXX17 MPI P4EST TRILINOS)
 
-if(NOT DEAL_II_WITH_MPI)
-  message(FATAL_ERROR "deal.II needs to be configured with MPI support.")
-endif()
+foreach(FEATURE ${DEAL_II_REQUIRED_FEATURES})
+  if(NOT DEAL_II_WITH_${FEATURE})
+    list(APPEND DEAL_II_MISSING_FEATURES ${FEATURE})
+    set(DEAL_II_HAS_ALL_DEPENDENCIES FALSE)
```
Rombur, line 11: This doesn't seem to be used

Rombur (approved):

PR #281, opened 2024-04-30
Updates for deal.II master past 9.5

masterleinad:
https://github.com/dealii/dealii/pull/16745 updated the SolverGMRES implementation and `deal.II/lac/la_vector.h` was removed in https://github.com/dealii/dealii/commit/d22e218017b8577437eb4fd36315dc6bf8653988.


Rombur:
I've thought about this before and I don't want to support multiples versions of our dependencies. We do it at one place but retrospectively this was a mistake because we never check that path. Supporting multiples versions of our dependencies makes maintaining and testing the code more complicated and we gain nothing from it.

masterleinad:
Superseded by #359.

Issue #190, opened 2022-10-17
GPU improvements (BCs, heat source, FE_Nothing)


stvdwtt:
- [ ] Matrix-free BCs on GPU
- [ ] Heat source on GPU
- [ ] FE_Nothing on GPU

stvdwtt:
Assigning the heat source part of this to @masterleinad

masterleinad:
The code that needs to be moved to the GPU is https://github.com/adamantine-sim/adamantine/blob/ba6699d8ddae57c3f3e7ca34e1b2db91a586cdc8/source/ThermalPhysics.templates.hh#L157-L175 The heat source classes are `HeatSource` (base class), `CubeHeatSource`, `ElectronBeamHeatSource`, and `GoldakHeatSource`. The position of the heat source is given by `ScanPath`.

Issue #223, opened 2023-02-21
Add capability to read updates to the scan path during a simulation


stvdwtt:
ROS will output something that looks like an "event series" format for adamantine to read.

stvdwtt:
From today's meeting the desired behavior is for adamantine to load the scan path file, run until the end of the commands in that file and then reparse the file to see if new lines have been appended. If they have then it simulates the appended scan path. If not, then it waits and then reparses the file.

stvdwtt:
Partially addressed by #291. The next step is to extend the scan path syntax to communicate whether more scan path updates are expected or not.

PR #282, opened 2024-05-03
[WIP] Introduce HeatSources

masterleinad:
This is a step towards https://github.com/adamantine-sim/adamantine/issues/190. This pull request replaces `std::vector<std::shared_ptr<adamantine::HeatSource<dim>>>` with `HeatSources` that contains all heat sources as Kokkos::Views and loops over them in its member functions. With this approach, there is no dynamic polymorphism and the compiler could inline all function calls for the heat sources. In particular, this object would be usable on GPUs.

The tests are still failing but I wanted to put this out already to see if there are objections to this approach.


source/ScanPath.cc
```diff
@@ -138,7 +140,9 @@ void ScanPath::load_event_series_scan_path(std::string scan_path_file)
     segment.power_modifier = last_power;
     last_power = std::stod(split_line[4]);
 
-    _segment_list.push_back(segment);
+    if (_segment_list_length == MAX_NUMBER_OF_SEGMENTS)
+      Kokkos::abort("too many segmnets");
```
stvdwtt, line 144: Typo:
```suggestion
      Kokkos::abort("too many segments");
```

source/ScanPath.hh
```diff
@@ -19,6 +19,8 @@
 #include <istream>
 #include <vector>
 
+#define MAX_NUMBER_OF_SEGMENTS 100
```
stvdwtt, line 22: This is too low -- if we had a 50-layer print with only two segments per layer, we'd hit this. I'm not sure that we've done enough sliced full parts to know a good upper bound. For example the AOP hourglass has 8055 segments and that's pretty simple as far as "real" parts go. Can we use a Kokkos::View with a runtime-defined length instead?

masterleinad: That was one of my next steps but I would need to manage the allocations for the segments outside the heat sources to avoid Views of managed Views. I'm not sure how much freedom we have to change those APIs.

The current version should be (more or less) device-usable already at least.

stvdwtt: Ok, got it. Maybe another option would be to make it a template parameter? Even with the dynamic scan path in #223, we wouldn't be changing the length of the scan path too many times per simulation.

masterleinad: Sure, changing it to a template argument for the heat source and the scan path is easy. The question then is if we want to have a runtime dispatch for pre-instantiated maximum sizes or any other mechanism that doesn't require rebuilding the library and the application for certain input files.

Rombur: We don't want to do this. We don't know how many segments we will need but it's potentially very large. If someone wants to point-source raster for PBF, the number of segments will be huge. 
In the current code, the scan path is part of the heat source. You need to change that. Move the scan path to `HeatSources`, add a function that evaluates the scan patch from `t_0` to `t_1` and fill a `Kokkos::View`. This way we don't need to know the number of segment and you don't waste memory if you have a ton a segments.

> I'm not sure how much freedom we have to change those APIs.

`adamantine` is not a library. Everything but the input parameters is an implementation details.

source/HeatSources.hh
```diff
@@ -0,0 +1,404 @@
...
+std::vector<ScanPath<MemorySpace>>
+HeatSources<MemorySpace, dim>::get_scan_paths() const
+{
+  std::vector<ScanPath<MemorySpace>> scan_paths;
+  for (unsigned int i = 0; i < _electron_beam_heat_sources.size(); ++i)
+    scan_paths.push_back(_electron_beam_heat_sources(i).get_scan_path());
+  for (unsigned int i = 0; i < _goldak_heat_sources.size(); ++i)
+    scan_paths.push_back(_goldak_heat_sources(i).get_scan_path());
+  return scan_paths;
+}
```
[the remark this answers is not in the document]
Rombur: No it's not. There is no order in the input file. You can reorder everything and the same input file will work.

source/ThermalPhysics.templates.hh
```diff
@@ -844,21 +809,9 @@ void ThermalPhysics<
...
-
-    // PropertyTreeInput sources.beam_X.type
-    std::string type = beam_database.get<std::string>("type");
-
-    if (type == "goldak" || type == "electron_beam")
-      source->set_beam_properties(beam_database);
-
-    source_index++;
-  }
+  auto host_heat_sources = _heat_sources.copy_to(dealii::MemorySpace::Host{});
```
Rombur, line 812:
```suggestion
  auto heat_sources_host = _heat_sources.copy_to(dealii::MemorySpace::Host{});
```

source/ThermalPhysics.templates.hh
```diff
@@ -844,21 +809,9 @@ void ThermalPhysics<
...
+  host_heat_sources.set_beam_properties(heat_source_database);
```
Rombur, line 813: We know that the database will always live on the host. It seems wasteful to copy everything back and forth.

source/ThermalPhysics.templates.hh
```diff
@@ -870,15 +823,8 @@ double ThermalPhysics<dim, p_order, fe_degree, MaterialStates, MemorySpaceType,
...
-  // maximum heat source height, which can lead to unexpected behavior for
-  // different sources with different heights.
-  double temp_height = std::numeric_limits<double>::lowest();
-  for (auto const &source : _heat_sources)
-  {
-    temp_height = std::max(temp_height, source->get_current_height(t));
-  }
-  _current_source_height = temp_height;
+  _current_source_height =
+      _heat_sources.copy_to(dealii::MemorySpace::Host{}).get_current_height(t);
```
Rombur, line 827: Same comment as above

source/ThermalPhysicsInterface.hh
```diff
@@ -19,9 +19,6 @@ namespace adamantine
 // Forward declarations
 class Timer;
 
-template <int dim>
-class HeatSource;
-
```
Rombur, line 24: You don't need a forward declaration because you are relying on the order of the includes.

source/ScanPath.hh
```diff
@@ -66,13 +69,16 @@ public:
...
+  ScanPath(
+      Kokkos::View<ScanPathSegment *, typename MemorySpaceType::kokkos_space,
+                   Kokkos::MemoryTraits<Kokkos::Unmanaged>>
+          scan_path_segments)
+      : _segment_list(scan_path_segments)
+  {
+  }
+
+  static std::vector<ScanPathSegment>
+  extract_scan_paths(std::string scan_path_file, std::string file_format);
```
Rombur, line 81: I don't really like this name. You are just reading from a file.

tests/test_heat_source.cc
```diff
@@ -29,8 +29,17 @@ BOOST_AUTO_TEST_CASE(heat_source_value_2d, *utf::tolerance(1e-12))
...
+      ScanPath<dealii::MemorySpace::Host>::extract_scan_paths(
+          database.get<std::string>("scan_path_file"),
+          database.get<std::string>("scan_path_file_format"));
+  Kokkos::View<ScanPathSegment *, Kokkos::HostSpace> scan_paths_segments_view(
+      scan_path_segments.data(), scan_path_segments.size());
+  BeamHeatSourceProperties beam(database);
+  GoldakHeatSource<2, dealii::MemorySpace::Host> goldak_heat_source(
+      beam, ScanPath<dealii::MemorySpace::Host>(scan_paths_segments_view));
+  ElectronBeamHeatSource<2, dealii::MemorySpace::Host> eb_heat_source(
+      database, ScanPath<dealii::MemorySpace::Host>(scan_paths_segments_view));
```
Rombur, line 42: Can you put this code in a function instead of copy/pasting three times.

tests/test_material_deposition.cc
```diff
@@ -353,7 +353,14 @@ BOOST_AUTO_TEST_CASE(material_deposition)
...
 {
-  adamantine::ScanPath scan_path("scan_path.txt", "segment");
+  std::vector<adamantine::ScanPathSegment> scan_path_segments =
+      adamantine::ScanPath<dealii::MemorySpace::Host>::extract_scan_paths(
+          "scan_path.txt", "segment");
+  Kokkos::View<adamantine::ScanPathSegment *, Kokkos::HostSpace>
+      scan_paths_segments_view(scan_path_segments.data(),
+                               scan_path_segments.size());
+  adamantine::ScanPath<dealii::MemorySpace::Host> scan_path(
+      scan_paths_segments_view);
```
Rombur, line 363: Put this code in a function to avoid the copy/paste.

Rombur (changes requested):
You will need to rebase the PR on develop. There was some small changes in `ScanPath`.

masterleinad:
Retest this please.

Rombur:
Let me know when you want another review on this.

masterleinad:
> Let me know when you want another review on this.

This could take another review now. Some notes:
- I create local copies on the host for the HeatSources object where necessary assuming that we want it by default to be in the memory space specified in the respective template argument for the other operators. It might make sense to keep it in host memory instead and only create device copies or we keep a member variable for both memory spaces (if they don't coincide).
- I needed to move a bunch of `const` members in the heat sources to make these classes copyable. I was running into some issues with `static` variables if used on the device as well. Thus, I moved the respective constants into member functions.
- The recent changes reusing scan file paths made the HeatSources code a little more complicated. We are now storing those `std::strings` in `HeatSources` directly as well and recreating the scan file segments from them if necessary. Since the `ScanPath` doesn't own the `ScanPathSegments`, reading the file is a static member instead of a member function named "extract_scan_paths". I'm happy to change it back to some other name like "read_file" again if preferred.
- I believe I addressed all other comments.

source/ElectronBeamHeatSource.hh
```diff
@@ -33,25 +34,82 @@ public:
...
+
+  /**
+   * Return the scan path.
+   */
+  ScanPath<MemorySpaceType> const &get_scan_path() const;
+
+  void set_scan_path(ScanPath<MemorySpaceType> const scan_path)
+  {
+    _scan_path = scan_path;
+  }
```
Rombur, line 59: Move the implementation out of the class declaration

source/ElectronBeamHeatSource.hh
```diff
@@ -33,25 +34,82 @@ public:
...
+
+  /**
+   * Compute the current height of the where the heat source meets the material
```
Rombur, line 62:
```suggestion
   * Compute the current height where the heat source meets the material
```

source/GoldakHeatSource.cc
```diff
@@ -9,29 +9,36 @@
...
 {
 }
 
-template <int dim>
-void GoldakHeatSource<dim>::update_time(double time)
+template <int dim, typename MemorySpaceType>
+void GoldakHeatSource<dim, MemorySpaceType>::update_time(double time)
 {
+  static const double pi_over_3_to_1p5 =
+      std::pow(dealii::numbers::PI / 3.0, 1.5);
```
Rombur, line 29: Why the change?

source/GoldakHeatSource.hh
```diff
@@ -33,25 +34,82 @@ public:
...
+
+  /**
+   * Return the scan path.
+   */
+  ScanPath<MemorySpaceType> const &get_scan_path() const;
+
+  void set_scan_path(ScanPath<MemorySpaceType> const scan_path)
+  {
+    _scan_path = scan_path;
+  }
```
Rombur, line 59: Move the implementation out of the declaration

source/HeatSources.hh
```diff
@@ -0,0 +1,482 @@
...
+   */
+  void update_time(double time);
+
+  /**
+   * Update the scan paths for all heat sources reading again from the
+   * respective files.
+   */
+  template <typename Dummy = MemorySpaceType>
+  std::enable_if_t<std::is_same_v<Dummy, dealii::MemorySpace::Host>>
+  update_scan_paths();
```
Rombur, line 52: `update_scan_paths()` should just do the right thing independent of the `MemorySpace`

source/HeatSources.hh
```diff
@@ -0,0 +1,482 @@
...
+  friend class HeatSources<dim, dealii::MemorySpace::Default>;
+  friend class HeatSources<dim, dealii::MemorySpace::Host>;
+
+  /**
+   * Private helper for update_scan_paths
+   */
+  void internal_update_scan_paths();
+
+  /**
+   * Private constructor used by copy_to_host.
```
Rombur, line 88: I cannot find `copy_to_host`

source/HeatSources.hh
```diff
@@ -0,0 +1,482 @@
...
+
+  std::vector<std::string> _goldak_scan_files;
+  std::vector<std::string> _electron_beam_scan_files;
+  std::vector<std::string> _goldak_scan_formats;
+  std::vector<std::string> _electron_beam_scan_formats;
+};
+
+template <int dim, typename MemorySpaceType>
+void HeatSources<dim, MemorySpaceType>::internal_update_scan_paths()
+{
```
Rombur, line 136: Put the implementation in a `.cc` not in the header.

application/adamantine.hh
```diff
@@ -1025,25 +1032,26 @@ run(MPI_Comm const &communicator, boost::property_tree::ptree const &database,
       }
       if (scan_path_updated)
       {
+        adamantine::HeatSources<dim, dealii::MemorySpace::Host>
+            heat_sources_host =
+                heat_sources.copy_to(dealii::MemorySpace::Host{});
+        heat_sources_host.update_scan_paths();
+        heat_sources = heat_sources_host.copy_to(MemorySpaceType{});
+
+        auto const &scan_paths = heat_sources.get_scan_paths();
```
Rombur, line 1041: This looks wrong. Shouldn't this be `heat_source_host.get_scan_path()`?

source/HeatSources.hh
```diff
@@ -0,0 +1,482 @@
...
+template <int dim, typename MemorySpaceType>
+template <typename Dummy>
+std::enable_if_t<std::is_same_v<Dummy, dealii::MemorySpace::Host>>
+HeatSources<dim, MemorySpaceType>::update_scan_paths()
+{
+  internal_update_scan_paths();
+
+  for (unsigned int i = 0; i < _goldak_heat_sources.size(); ++i)
+    _goldak_heat_sources(i).set_scan_path(
+        ScanPath<MemorySpaceType>{_goldak_scan_path_segments[i]});
```
Rombur, line 178: Why are we keeping track of the scan path in `HeatSources` and in each individual sources? We should not duplicate this information.

source/HeatSources.hh
```diff
@@ -0,0 +1,455 @@
...
+                                                double const height) const
+{
+  double value = 0;
+  for (unsigned int i = 0; i < _electron_beam_heat_sources.size(); ++i)
+    value += _electron_beam_heat_sources(i).value(point, height);
+  for (unsigned int i = 0; i < _cube_heat_sources.size(); ++i)
+    value += _cube_heat_sources(i).value(point, height);
+  for (unsigned int i = 0; i < _goldak_heat_sources.size(); ++i)
+    value += _goldak_heat_sources(i).value(point, height);
+  return value;
```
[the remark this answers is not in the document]
Rombur: ?

source/ScanPath.cc
```diff
@@ -8,41 +8,45 @@
...
 #include <fstream>
 
 namespace adamantine
 {
-ScanPath::ScanPath(std::string scan_path_file, std::string file_format)
-    : _scan_path_file(scan_path_file), _file_format(file_format)
+
+template <typename MemorySpaceType>
+std::vector<ScanPathSegment>
+ScanPath<MemorySpaceType>::extract_scan_paths(std::string scan_path_file,
```
Rombur, line 22: I still hate the name

source/ScanPath.hh
```diff
@@ -54,6 +54,7 @@ class ScanPathTester;
  * gives the power modifier for the current segment. It reads in the scan path
  * from a text file.
  */
+template <typename MemorySpaceType>
```
Rombur, line 57: Is is actually necessary to have the ScanPath on the GPU? I feel like this should only exists on the CPU and we should only move part of the data when it's necessary. Otherwise, we will end up using several GBs of memory on the GPU just for the ScanPath.

source/ThermalPhysicsInterface.hh
Rombur: You still need a forward declaration, otherwise the code may break if the include are reordered.

PR #286, opened 2024-05-17
Draft JOSS paper

Rombur:
@stvdwtt Here is the draft of the JOSS paper. We can either iterate on the PR or work on the branch directly. It's up to you. Here are the requirements for the paper: https://joss.readthedocs.io/en/latest/paper.html


Rombur:
The draft is in pretty good shape. I am merging it

PR #287, opened 2024-05-20
Avoid compiler warnings

masterleinad:
While working on #282, I noticed that there are more warnings which this pull request fixes. I wasn't sure if we wanted to enforce this with `-Werror` or similar.

### DIFF
--- a/application/adamantine.cc
+++ b/application/adamantine.cc
@@ -106,7 +106,7 @@ get_p_order_and_n_material_states(boost::property_tree::ptree &database)
 
   // Sanity check
   adamantine::ASSERT_THROW(
-      p_order >= 0 && p_order < 5,
+      p_order < 5,
       "Error when computing the polynomial order of the material properties");
   adamantine::ASSERT_THROW(
       n_material_states > 0 && n_material_states < 4,
--- a/source/MaterialProperty.hh
+++ b/source/MaterialProperty.hh
@@ -57,13 +57,6 @@ public:
       boost::property_tree::ptree const &database);
 
   /**
-   * Copy constructor. It should only be called by KOKKOS_CLASS_LAMBDA. The
-   * copy constructor does not copy all the member variables of the class.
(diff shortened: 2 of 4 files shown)


stvdwtt (approved):
These look good to me. @Rombur, any concerns?

ci/Dockerfile
```diff
@@ -220,6 +220,7 @@ RUN export TRILINOS_HASH=975307431d60d0859ebaa27c9169cbb1d4287513 && \
         -DTrilinos_ENABLE_SEACAS=ON \
         -DTrilinos_ENABLE_EXPLICIT_INSTANTIATION=ON \
         -DKokkos_ENABLE_CUDA=$CUDA \
+        -DKokkos_ENABLE_CUDA_CONSTEXPR=$CUDA \
```
Rombur, line 223: Don't modify the Dockerfile. This will unsync the image with the Dockerfile.

Rombur (changes requested):

ci/Dockerfile
masterleinad: OK. I'll remove but most of the warnings are related to setting this option.

Rombur: Then why are you setting this option if it creates a bunch of warnings?

masterleinad: It removes all warnings related to calling `constexpr` `host` functions. This is also what we are doing in the `deal.II` CI, see https://github.com/dealii/dealii/blob/3c3910b004cec9be552eb224ceba4bc6fecd47b6/.github/workflows/linux.yml#L335, and we check for this option to be set for an external standalone `Kokkos`, see https://github.com/dealii/dealii/blob/3c3910b004cec9be552eb224ceba4bc6fecd47b6/cmake/modules/FindDEAL_II_KOKKOS.cmake#L53.

Rombur (approved):

PR #288, opened 2024-05-23
Fix uninitialized variables

Rombur:
I found the cause of the random failings in the CI. In some tests, we forgot to set the time at which the source should be evaluated. I have initialized member data in the sources to NaN which allowed me to find errors in `test_thermal_operator_device` and `test_implicit_operator` on top  of the known issue with `test_thermal_operator`.

### DIFF
--- a/tests/test_thermal_operator.cc
+++ b/tests/test_thermal_operator.cc
@@ -91,13 +91,14 @@ BOOST_AUTO_TEST_CASE(thermal_operator, *utf::tolerance(1e-15))
   beam_database.put("depth", 0.1);
   beam_database.put("absorption_efficiency", 0.1);
   beam_database.put("diameter", 1.0);
-  beam_database.put("max_power", 10.);
+  beam_database.put("max_power", 0.);
   beam_database.put("scan_path_file", "scan_path.txt");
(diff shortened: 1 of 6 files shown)


masterleinad (approved):

tests/test_thermal_operator.cc
```diff
@@ -91,13 +91,14 @@ BOOST_AUTO_TEST_CASE(thermal_operator, *utf::tolerance(1e-15))
   beam_database.put("depth", 0.1);
   beam_database.put("absorption_efficiency", 0.1);
   beam_database.put("diameter", 1.0);
-  beam_database.put("max_power", 10.);
```
stvdwtt, line 94: What's your reason for setting these to zero? That the test doesn't actual evaluate the source term?

Rombur: These tests compare the matrix-free implementation with matrix-based implementation by doing a matrix-vector multiplication. For the matrix-based implementation, we don't use the rhs, i.e., the source term. For the matrix-free implementation, the evaluation of the source term is embedded in the evaluation of the operator, so we need to put the source to zero to match the matrix-based implementation.

stvdwtt (approved):

PR #291, opened 2024-05-30
Add support for the on-the-fly update of the scan path

Rombur:
@stvdwtt can you try this PR. With it you shouldn't need to restart adamantine to update the scan path.

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -37,6 +37,7 @@
 #include <deal.II/lac/vector_operation.h>
 #include <deal.II/numerics/error_estimator.h>
 
+#include <boost/algorithm/string.hpp>
 #include <boost/archive/text_iarchive.hpp>
 #include <boost/archive/text_oarchive.hpp>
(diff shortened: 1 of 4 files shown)


stvdwtt:
This works for expected with the scan path: if I swap the scan path file before the simulation gets to the end of the initial scan path file it will load the updated scan path and do the full scan path.

One thing before I approve and merge -- are you intentionally saving recalculation of material addition for a second PR? Because it looks like only the material from the initial scan path gets deposited.

Rombur:
> One thing before I approve and merge -- are you intentionally saving recalculation of material addition for a second PR? Because it looks like only the material from the initial scan path gets deposited.

No,  it's a copy/paste error. Can you remove the `auto` at line 1043 of adamantine.hh and try again

Rombur:
Sorry replace the `auto` by `std::tie(...)`

stvdwtt (approved):
This has the expected behavior now -- approving and merging.

PR #293, opened 2024-06-17
Restart fixes for experimental data and ensembles

stvdwtt:
This PR makes two small fixes involving restart:
- It removes the rank id from checkpoint filenames, this isn't needed and prevents loading the checkpoint at restart.
- It advances the experimental frame as necessary upon restart. Before this PR the code aborts when restarting after an experimental frame because it identifies an old, unassimilated frame.

I had to move the experiment block to after the restart block in initialization, this shows up in git as a somewhat convoluted change.

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -1482,10 +1482,8 @@ run_ensemble(MPI_Comm const &global_communicator,
     // PropertyTreeInput checkpoint.time_steps_between_checkpoint
     time_steps_checkpoint =
         checkpoint_database.get<unsigned int>("time_steps_between_checkpoint");
-    // PropertyTreeInput checkpoint.filename_prefix
     checkpoint_filename =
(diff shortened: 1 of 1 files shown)


application/adamantine.hh
```diff
@@ -1482,10 +1482,8 @@ run_ensemble(MPI_Comm const &global_communicator,
     // PropertyTreeInput checkpoint.time_steps_between_checkpoint
     time_steps_checkpoint =
         checkpoint_database.get<unsigned int>("time_steps_between_checkpoint");
-    // PropertyTreeInput checkpoint.filename_prefix
```
Rombur, line 1485: Keep the comment

Rombur (approved):
Looks good. I think you remove a comment by error?

application/adamantine.hh
stvdwtt: Good catch -- I added that back in.

Issue #272, opened 2024-04-16
Modify how simulations end


stvdwtt:
Currently the simulation runs until a designated end time. We discussed changing this so that the simulation end time is optionally dictated by the scan path. This requires extending the syntax of the scan path files to include something that denotes the end of the simulation. (This is needed for dynamic changes to the scan path in #223 to disambiguate between a simulation that is waiting for more scan path info and a simulation ready to be completed.)

stvdwtt:
Fixed in #294

PR #294, opened 2024-06-17
Improve support for scan path update

### LINKED ISSUE #223: Add capability to read updates to the scan path during a simulation

Rombur:
With this PR, we stop reading the scan path when we read `SCAN_PATH_END`. When using `segment scan path`, we need the number of segments to be at least equal to number of  segments + 1. Otherwise we won't read the `SCAN_PATH_END` line.

@stvdwtt can you check that it works as you expect.

Fixes #223  and #272


stvdwtt:
Works like a charm!

PR #301, opened 2024-07-17
Fix mechanical simulation with MPI

masterleinad:
This fixes running thermo-mechanical simulations with multiple MPI processes (without adaptivity). The most important changes are due to the mesh update in `add_material`. We need to use `SolutionTransfer`/`CellDataTransfer` for the mechanical data as well since activating elements shifts elements to different MPI processes. The `SolutionTransfer`/`CellDataTransfer` objects need to be used in order which means that we need to split `add_material` in a part that sets up the triangulation for p-update, performing the update, and then interpolating to the new mesh.
We activate elements for the mechanical simulation later separately but we could possibly do that at the time. For now, it's fine to update the finite elements used in that step without communicating that information across all processes (which might shift elements across processes again and require transferring data for the thermal simulation).
Using `CellDataTransfer` in particular requires us to provide data for all active cells, not just all locally owned cells.

The changes are tested by running `test_integration_thermoelastic` with multiple MPI processes.

PR #328, opened 2024-10-14
Add ClosestSourcePointAdaptation

masterleinad:
Related to #301. This pull request introduces a class that handles data transfer in CellDataTransfer when cells are coarsened or refined by assigning the value at the closest quadrature point.

### DIFF
--- a/source/ClosestSourcePointAdaptation.hh
+++ b/source/ClosestSourcePointAdaptation.hh
@@ -1,0 +1,117 @@
+/* Copyright (c) 2016 - 2024, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
+
+#ifndef CLOSEST_SOURCE_POINT_ADAPTATION_HH
+#define CLOSEST_SOURCE_POINT_ADAPTATION_HH
+
+#include <deal.II/fe/fe_nothing.h>
+#include <deal.II/fe/fe_values.h>
+
+#include <random>
+
+namespace adamantine
+{
+
+template <int dim, int spacedim, typename quad_point_value_type>
+class ClosestQuadPointAdaptation
+{
+public:
+  template <class Quadrature>
+  ClosestQuadPointAdaptation(const Quadrature &quadrature)
+      : m_fe_values_coarse(m_fe_nothing, quadrature,
+                           dealii::update_quadrature_points),
+        m_fe_values_fine(m_fe_nothing, quadrature,
+                         dealii::update_quadrature_points)
+  {
+  }
+
+  std::vector<std::vector<quad_point_value_type>> coarse_to_fine(
+      const typename dealii::Triangulation<dim, spacedim>::cell_iterator
+          &parent,
+      const std::vector<quad_point_value_type> parent_values)
+  {
+    m_fe_values_coarse.reinit(parent);
+    const std::vector<dealii::Point<spacedim>> &coarse_quad_points =
+        m_fe_values_coarse.get_quadrature_points();
+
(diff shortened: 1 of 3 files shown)


Rombur:
@masterleinad Do you remember why we had this PR? Is it a step toward supporting AMR with the mechanics?  It would be nice if the entire code could use AMR.

masterleinad:
> @masterleinad Do you remember why we had this PR? Is it a step toward supporting AMR with the mechanics? It would be nice if the entire code could use AMR.

Yes, exactly.

masterleinad:
The tricky part is to properly register CellDataTransfer with grid refinement/coarsening but as a standalone feature this should be fine.

Rombur:
OK great. I'll take a look at it

source/ClosestSourcePointAdaptation.hh
```diff
@@ -0,0 +1,117 @@
+/* Copyright (c) 2016 - 2024, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
```
Rombur: We have changed the license since the PR was opened

tests/test_closest_source_point_adaptation.cc
```diff
@@ -0,0 +1,129 @@
+/* Copyright (c) 2016 - 2024, the adamantine authors.
+ *
+ * This file is subject to the Modified BSD License and may not be distributed
+ * without copyright and license information. Please refer to the file LICENSE
+ * for the text and further information on this license.
+ */
```
Rombur, line 3: Wrong license

tests/test_closest_source_point_adaptation.cc
```diff
@@ -0,0 +1,129 @@
...
+      n_active_cells,
+      std::vector<double>(n_q_points,
+                          std::numeric_limits<double>::signaling_NaN()));
+
+  cell_data_transfer.unpack(cell_quad_point_values_final);
+
+  for (auto &cell : tria.active_cell_iterators())
+    if (cell->is_locally_owned())
+    {
+      cell->set_refine_flag();
```
Rombur: You don't need this

Rombur (commented):
The PR looks fine but it needs more documentation. People who don't know deal.II will have a hard time understand what's going on and why we do it.

tests/test_closest_source_point_adaptation.cc
```diff
@@ -0,0 +1,128 @@
...
+#include <deal.II/fe/fe_values.h>
+#include <deal.II/grid/grid_generator.h>
+
+#include <random>
+#include <string>
+
+#include "main.cc"
+
+// Check that ClosestSourcePointAdaption can be used with dealii::CellDataTransfer
+// In particular, check that we get obtain the same qudratire point values when we first refine and then coarsen the mesh.
```
Rombur:
```suggestion
// In particular, check that we get obtain the same quadrature point values when we first refine and then coarsen the mesh.
```

source/ClosestSourcePointAdaptation.hh
```diff
@@ -0,0 +1,134 @@
...
+#include <random>
+
+namespace adamantine
+{
+
+/**
+ * This class is supposed to be used with dealii::CellDataTransfer to
+ *interpolate quadrature point values when coarsening and refining the mesh. The
+ *approach taken by this class is to use the value stored in the closest
+ *previous quadrature point.
```
Rombur:
```suggestion
 * interpolate quadrature point values when coarsening and refining the mesh. The
 * approach taken by this class is to use the value stored in the closest
 * previous quadrature point.
```

Rombur (approved):

PR #333, opened 2024-10-28
Fix setup of plastic variables

Rombur:
@stvdwtt Can you test this PR.


stvdwtt:
@Rombur, the fix in this PR now gives the intended result for a serial simulation. There is however still an issue for multiple MPI processes.

My recommendation is to merge this PR and then we can debug the MPI issue seperately.

stvdwtt (approved):

masterleinad (approved):

PR #350, opened 2024-12-17
Fix Doxygen warnings

Rombur:
This PR fixes Doxygen and stop printing full path of the directory where the Doxygen documentation was built.


Rombur:
The changes are only in Doxyfile.in and in documentation, so we don't need to run the CI. Since I need this PR for something else, I am merging as is.

PR #353, opened 2024-12-26
Rework ensembles

Rombur:
This PR does three things:
 - Make sure that the variables are not correlated. Currently, if we call the function `get_normal_random_vector` twice in a row, we get the same values. This is because each call builds a new distribution using the same seed. Fix this issue by rejecting an increasing number of initial draws.
 - Reject draws that return a negative value which are non-physical.  I am not sure if this is better or if we should just error out.
 - For each ensemble member write a file with the data they used.


stvdwtt (approved):

stvdwtt:
This looks good to me -- merging now.

PR #356, opened 2025-01-07
Use ArborX to find which cells to refine when using AMR

Rombur:
There are two problems with the way AMR currently works:
 1. each member refines its own mesh and therefore we cannot guarantee that all the members use the exact same mesh
 2. the way we flag the cells for refinement is very inefficient. For each time step, we loop over the entire mesh to build a list of the cells that we need to refine.

To fix problem 1., we build bounding sources that bound the sources of all the ensemble members. We use these common sources to determine which cells need to be refined.

To fix problem 2., we use ArborX to compute the intersections of the bounding sources with the mesh. I decided to accumulated all the bounding sources before calling ArborX. This is to avoid calling ArborX with a single bounding box. In my tests, this works well if the number of time steps between refinement is in the order of 100 000 or lower. For higher number of time steps, building the bounding boxes requires quite a bit of memory. If it's a problem in practice, we can revisit this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,6 @@ The following options are available:
   * additional\_output\_refinement: additional levels of refinement for the output (default: 0)
 * refinement (required):
   * n\_refinements: number of times the cells on the paths of the beams are refined (default value: 2)
-  * beam\_cutoff: the cutoff value of the heat source terms above which beam-based refinement occurs (default value: 1e-15)
   * coarsen\_after\_beam: whether to coarsen cells where the beam has already passed (default value: false)
   * time\_steps\_between\_refinement: number of time steps after which the
   refinement process is performed (default value: 2)
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -543,6 +543,7 @@ compute_cells_to_refine(
   }
   dealii::ArborXWrappers::BVH bvh(cell_bounding_boxes);
 
+  double const bounding_box_scaling = 2.0;
   std::vector<dealii::BoundingBox<dim>> heat_source_bounding_boxes;
   for (unsigned int i = 0; i < n_time_steps; ++i)
   {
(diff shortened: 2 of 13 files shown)


stvdwtt:
Do you have any examples for how much the fix to Problem 2 decreases the time spent on remeshing?

Rombur:
It depends on how many time steps you do before remeshing. If you use 50 000 or more, the entire remeshing time is spent just flagging the cells. You don't touch the mesh very often but you need to go through the mesh 50 000 times. I don't have the timing anymore but if I remember correctly, you can't really use such a large number with the current master. It takes longer than evaluating the operator because on each cells you need to ask for the face, then the center of the face, then you can evaluate heat source. That's just super slow. Now we need to pay to build the bounding boxes for the cells but we can reuse them for multiple times steps and we don't need to do comparison with the entire mesh.

stvdwtt:
@Rombur generally this looks good to me. My one suggestion would be to add a user option to scale the bounding box. I'm not sure a priori where the melt pool bounds are compared to nominal diameter and depth of the heat source.

Rombur:
> My one suggestion would be to add a user option to scale the bounding box. I'm not sure a priori where the melt pool bounds are compared to nominal diameter and depth of the heat source.

We already have so many parameters. Would you be fine if I just hard code a scaling factor for now? If we realize that it is important to pick the right parameter depending on the problem, then I'll add an entry in the input file.

> The old criterion of the value being below 1e-15 contains much more volume than the nominal width and depth.

For the depth, the criterion is the same because the value of the source is assumed to be zero when the source penetration is bigger than depth. It is very different for the width though.

stvdwtt:
I think hard-coding a 2x factor on the diameter is a reasonable place to start and then we can either modify that hard-coded value or add an input parameter based on how well that works.

Rombur:
This is ready to be merged. I have applied the scaling factor to all the dimensions since we don't really know what the volume should look like.

Issue #357, opened 2025-01-10
How to impose temperature dependent material properties?


wd15:
Are temperature dependent material properties possible? If so, how does one write them in the `.info` file.

Rombur:
Yes, it is supported. You can either describe the property variation as a polynomial or as a table (I haven't used the table as much though). You need to set the [property_format](https://adamantine-sim.github.io/adamantine/doc/input_file.html#materials-required) to `table` or `polynomial`. 
Here is an example of the input using the polynomial format: https://github.com/adamantine-sim/adamantine/blob/5d54d9771382ec5f875e3b968139615883311240/tests/test_material_property.hh#L416-L429
You give the coefficient for order 0, then 1, etc. You don't need to use the same order for all material properties.
Here is an example of the input using the table format: https://github.com/adamantine-sim/adamantine/blob/5d54d9771382ec5f875e3b968139615883311240/tests/test_material_property.hh#L299-L313
You give the temperature and  the value at this temperature. The temperature needs to be given in increasing value. We interpolate the value between the one you give us.

wd15:
Thanks for the response. The following gives an error.

```
materials
{
  n_materials 1
  property_format table
  
  material_0
  {
    solid
    {
      density 300.0, 7904; 2700.0, 7904
      specific_heat 300.0, 714; 2700.0, 714
      thermal_conductivity_x 300.0, 7.2; 2700.0, 42.0
      thermal_conductivity_y 300.0, 7.2; 2700.0, 42.0
      thermal_conductivity_z 300.0, 7.2; 2700.0, 42.0
    }

    powder
    {
      specific_heat 300.0, 714; 2700.0, 714
      density 300.0, 7904; 2700.0, 7904
      thermal_conductivity_x 300.0, 0.314; 2700.0, 0.314
      thermal_conductivity_y 300.0, 0.314; 2700.0, 0.314
      thermal_conductivity_z 300.0, 0.314; 2700.0, 0.314
    }

    liquid
    {
      specific_heat 300.0, 847; 2700, 847
      density 300.0, 7904; 2700.0, 7904
      thermal_conductivity_x 300.0, 37.3; 2700.0, 37.3
      thermal_conductivity_y 300.0, 37.3; 2700.0, 37.3
      thermal_conductivity_z 300.0, 37.3; 2700.0, 37.3
    }

    solidus 1675
    liquidus 1708
    latent_heat 290000
  }
}
```

The error is

```                                                    
[...]

PR #358, opened 2025-01-13
Fix input file when using temperature-dependent materiel properties

Rombur:
This PR is related to https://github.com/adamantine-sim/adamantine/issues/357
 - Do not use `;` as a delimiter when reading material properties because Boost thinks it is a comment. 
 - Rewrite the tests to read from file instead of building a property tree in code. 
 - Improve documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ The following options are available:
     * width\_divisions: number of cell layers in width (only in 3D) (default value: 10)
 * materials (required):
   * n\_materials: number of materials (required)
-  * property\_format: format of the material property: table or polynomial (required)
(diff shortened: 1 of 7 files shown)


wd15:
Any reason why this value needs to be 4? Could it be something more reasonable like 12?

https://github.com/adamantine-sim/adamantine/blob/5d54d9771382ec5f875e3b968139615883311240/source/MaterialProperty.hh#L46

Rombur:
> Any reason why this value needs to be 4? Could it be something more reasonable like 12?

I'll change it. I only use the polynomial format with at most 4 coefficients that's the only reason I picked 4 for the table size.

Rombur:
I've set the table size to 12. All the tests are passing. So I am merging

PR #359, opened 2025-01-17
Update software stack

Rombur:
This PR updates the software stack from deal.II 9.5 to deal.II 9.6. There are some minimal changes in the code associated with the update.

@wd15 Can you take a look at the flake. It works on my machine but maybe I am missing something. I did three changes on the flake:
 1. I updated nixos from 24.05 to 24.11
 2. I removed the dependency on CUDA. The current flake does not work with CUDA because we need to use `nvcc_wrapper` from Kokkos to compile most of the stack not the default compilers
 3. I have added deal.II 9.6 to the flake

I will need to retag our Docker images before we can merge this PR.


wd15:
FYI, all three Nix configurations run some examples. Probably should add a github action for the Nix build.  Something like [this](https://github.com/usnistgov/fipy/blob/master/.github/workflows/nix.yml) would work. I'll submit a PR for that at some point.

Issue #141, opened 2022-02-11
Re-group ensemble and data assimilation inputs


stvdwtt:
Currently, the input parameters controlling the standard deviation of model parameters for ensembles and whether or not to include them in augmented states are grouped in the "ensemble" and "data assimilation" input blocks. This is not desirable since it splits up where nominal parameter values are defined and where other attributes of that parameter are defined. Therefore, these should be moved to the appropriate "beam" or "material" blocks. It also simplifies setting these ensemble/data assimilation parameters for multiple heat sources and/or materials.

Along with this, we should implement a more flexible and scalable approach for parsing and handling the augmentation parameters. The `enum class AugementedStateParameters` in `DataAssimilator.hh` is heavy-handed, and promotes specifically enumerating which properties (of which beams/materials) can be used in augmented ensembles.

PR #361, opened 2025-01-30
Rework data assimilation variables

Rombur:
This PR reworks how we set variables for data assimilation. Currently the variables that we can set are hard-coded, for instance `beam_0_max_power_stddev` sets the standard deviation for the maximum power of `beam_0`. Instead of hard-coding the values, this PR rework how we can set the standard deviation. With this PR, `beam_0_max_power_stddev` is replaced by `sources.beam_0.max_power_stddev`. The idea is to follow the naming scheme of the variable and to add `_stddev` at the end of the name. We then check that the original variable exists. Since nothing is hard-coded, more variables can be used for data assimilation. Note however that data assimilation only works for scalar values and therefore it does not work for temperature dependent variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,10 +224,7 @@ The following options are available:
 * ensemble (optional):
   * ensemble\_simulation: whether to perform an ensemble of simulations (default value: false)
   * ensemble\_size: the number of ensemble members for the ensemble Kalman filter (EnKF) (default value: 5)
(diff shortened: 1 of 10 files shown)


stvdwtt (approved):
Note that this is related to #141. I think I disagree with what I said in there that the stddev variables should be spread throughout the input file with the nominal values. I think the current approach in this PR works well.

PR #366, opened 2025-02-09
Improved directory handling

cadkin:
Split out of #364.

Allows for the explicit definition of the output directory via a '-o' flag on the command line. The quick implementation sticks the output dir in the input file ptree, so this has (unintended) side-effect of allowing configuration of the output directory from the .info file. The key is post_processing.output_dir.

Additionally, also makes a small change where if you call adamantine from a different directory than the .info file, it will change the PWD to that containing directory. This fixes a small bug where adamantine would just spin its heels waiting on files in the wrong locations.

### DIFF
--- a/application/adamantine.cc
+++ b/application/adamantine.cc
@@ -146,7 +146,10 @@ int main(int argc, char *argv[])
     boost_po::options_description description("Options:");
     description.add_options()("help,h", "Produce help message.")(
         "input-file,i", boost_po::value<std::string>(),
-        "Name of the input file.");
+        "Name of the input file.")("output-dir,o",
+                                   boost_po::value<std::string>(),
+                                   "Output directory; defaults to the current working directory.");
+
     // Declare a map that will contains the values read. Parse the command
     // line and finally populate the map.
     boost_po::variables_map map;
@@ -196,6 +199,30 @@ int main(int argc, char *argv[])
       std::cerr << exception.what() << std::endl;
       return 0;
     }
+
+    // Abusing the database to pull out the output dir when writing the output
+    if (map.count("output-dir") == 1)
+    {
+      try
+      {
+        std::string outdir =
+            std::filesystem::absolute(map["output-dir"].as<std::string>())
+                .string() +
+            std::filesystem::path::preferred_separator;
+        std::filesystem::create_directories(outdir);
+
+        database.put("post_processor.output_dir", outdir);
+      }
+      catch (std::runtime_error const &exception)
+      {
+        std::cerr << exception.what() << std::endl;
+        return 0;
+      }
+    }
+
+    // Make adamantine behave a bit better and not lock up if not in cwd of
+    // current input file
+    std::filesystem::current_path(std::filesystem::absolute(filename).parent_path());
 
 #ifdef ADAMANTINE_WITH_CALIPER
     cali::ConfigManager caliper_manager;
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -14,14 +14,14 @@
   ],
   "buildPresets": [
     {
-      "name":            "debug-gcc-ninja",
+      "name":            "build-debug-gcc-ninja",
       "displayName":     "Debug (gcc-ninja)",
       "description":     "Compile a debug build using GCC.",
       "configurePreset": "generic-gcc-ninja",
       "configuration":   "Debug"
     },
     {
-      "name":            "release-gcc-ninja",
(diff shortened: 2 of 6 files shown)


application/adamantine.cc
```diff
@@ -197,6 +200,35 @@ int main(int argc, char *argv[])
...
+
+        database.put("post_processor.output_dir", outdir);
+      }
+      catch (std::runtime_error const &exception)
+      {
+        std::cerr << exception.what() << std::endl;
+        return 0;
+      }
+    }
+    else if (database.count("post_processor.output_dir") == 0)
```
Rombur, line 222: Remove the `else if` and instead of using `post_processor_database.get<std::string>("output_dir");` use `post_processor_database.get<std::string>("output_dir","");` Otherwise you will need to go in all the tests that use `PostProcessor` and add `output_dir` in their database.

application/adamantine.cc
```diff
@@ -197,6 +200,35 @@ int main(int argc, char *argv[])
...
+    {
+      database.put("post_processor.output_dir", "");
+    }
+
+    // Make adamantine behave a bit better and not lock up if not in cwd of
+    // current input file
+    std::filesystem::current_path(
+        std::filesystem::path(filename).parent_path());
```
Rombur, line 230: @stvdwtt can you check that the workflow code and the INTERSECT adapters still work with this change.

Rombur: @stvdwtt any update on this?

stvdwtt: @cadkin, @Rombur -- sorry for the delay. I'll do some testing with the control driver by the end of the week to make sure it doesn't cause issues (or issues that are easily addressed).

Rombur: @stvdwtt anything new on this?

application/adamantine.cc
```diff
@@ -146,7 +146,10 @@ int main(int argc, char *argv[])
     boost_po::options_description description("Options:");
     description.add_options()("help,h", "Produce help message.")(
         "input-file,i", boost_po::value<std::string>(),
-        "Name of the input file.");
+        "Name of the input file.")("output-dir,o",
+                                   boost_po::value<std::string>(),
+                                   "Output directory; defaults to $PWD.");
```
stvdwtt, line 151: Is this a typo?
```suggestion
                                   "Output directory; defaults to $CWD.");
```

application/adamantine.cc
stvdwtt: Sorry @cadkin and @Rombur for taking so long on this. I finally tested it with the driver. I was able to modify the driver to work with this change pretty straightforwardly.

Unless I'm missing something, it appears that we can no longer give the input file as just the name of the file, it has to be a path. Before the standard approach was to have the input file in the CWD and point to the input file with `-i input.info`. Now, I believe it has to be `-i ./input.info`. 

@cadkin, do you know if there's any simple way to support the older syntax without the `./`? If not, we'd have to update the documentation as part of this PR.

Rombur: > do you know if there's any simple way to support the older syntax without the ./?

If it's not possible, we could add a new option maybe something like `-p` (`-input_path`) with the new behavior.

stvdwtt: @Rombur, that's a good idea. I'm checking right now to see if it's possible to autodetect a relative path and treat it differently, but if that doesn't work I like adding a new option

stvdwtt:
```suggestion
        if (std::filesystem::path(filename).has_parent_path()){
          std::filesystem::path(filename).parent_path());
        }
```

This seems to have the intended effect of only changing the CWD if `filename` has a parent path.

@cadkin, would this work for you?

application/adamantine.cc
cadkin: I think in my brain I was referring to `$PWD` the environment variable since that's what `cwd` is in a shell script.

Anyway, I just changed it such that it actually spells it out so there's no ambiguity.

application/adamantine.cc
cadkin: Interesting, I would have expected that `std::filesystem::path` would have treated a path the same regardless of how it was input... very strange.

It seems that using `std::filesystem::absolute` instead forces the `parent_path()` to be defined since it can't be missing in an absolute path. Give my latest commit a try and let me know!

stvdwtt: This works! Approving and merging now.

stvdwtt (approved):

PR #369, opened 2025-02-14
testing flags

AshGannon:


tests/test_integration_3d.cc
```diff
@@ -92,6 +92,84 @@ BOOST_AUTO_TEST_CASE(integration_3D, *utf::tolerance(0.1))
     }
   }
 }
+/////////////////
+BOOST_AUTO_TEST_CASE(integration_3D, *utf::tolerance(0.1))
```
Rombur, line 96: You cannot reuse `integration_3D`. It needs to be a unique name. Maybe use sth like `integration_3D_demo_316`. You can just create a new commit and push on the branch. There is no need to close the PR and open a new one.

Rombur: Never mind close the PR and reopen a new one using the correct branch.

tests/test_integration_3d.cc
```diff
@@ -93,6 +93,88 @@ BOOST_AUTO_TEST_CASE(integration_3D, *utf::tolerance(0.1))
...
+  int num_ranks = 0;
+  MPI_Comm_size(communicator, &num_ranks);
+
+  // Limits for a weak non-pointwise check
+  double max_expected = 580.0;
+  double min_expected = 285.0;
+
+  if (num_ranks == 1)
+  {
+    //Write the file for rank 1
```
Rombur, line 123: This is wrong. You want to use this code once to create the gold file. Once you have it, you don't want to update it. You should only read it and compare the new result to it. Here you always overwrite the file and so there is no way to know if the values change over time.

tests/test_integration_3d.cc
```diff
@@ -93,6 +93,88 @@ BOOST_AUTO_TEST_CASE(integration_3D, *utf::tolerance(0.1))
...
+      gold_file >> gold_value;
+      BOOST_TEST(temperature.local_element(i) == gold_value);
+    }
+  }
+  else if (num_ranks == 2)
+  {
+    int rank = 0;
+    MPI_Comm_rank(communicator, &rank);
+
+    // To write the gold file
```
Rombur, line 146: Same problem as above

tests/test_integration_3d.cc
AshGannon: This is the first time it is running correctly. I need to swap the flags and comment these sections out and push it again. (I lied about it being the last time.)

AshGannon:
Closing in favor of #371

PR #370, opened 2025-02-17
Removing ffast-math from docker ci

AshGannon:


.github/workflows/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN cd /home && git clone https://github.com/adamantine-sim/adamantine && \
     cmake \
       -DDEAL_II_DIR=${DEAL_II_DIR} \
       -DCMAKE_BUILD_TYPE=Release\
-      -DCMAKE_CXX_FLAGS="-g -ffast-math" \
+      -DCMAKE_CXX_FLAGS="-g" \
```
Rombur, line 24: Actually can you remove this line entirely. Otherwise I fear that we won't have any optimization flag.

AshGannon: Amended

Rombur (approved):

Rombur:
Not waiting for the CI, since this is not tested by the CI.

PR #375, opened 2025-02-28
Improve ray tracing support for non axis aligned cells

Rombur:
This PR reworks the ray tracing. Currently we assume that the cells are axis aligned. This PR removes this assumption. The big change is that we need to do a proper fine search to find out if the ray intersects the cells. To do that I've split each face of the candidate cells into triangle and then check if the ray intersects a triangle. To determine if a point belongs to a triangle, we calculate its barycentric coordinates. If they are positive, the point is in the triangle. Otherwise it is outside.


stvdwtt:
@Rombur, have you checked this with real IR data to see if we get more intersections?

Rombur:
No, I haven't. I am not sure that we will get more intersections because currently we assume that the cells are the same as  the bounding boxes. If a ray hits a bounding box, we assume it hits the cells and then, we just assign the temperature to the closest quadrature point of the cell. This PR fixes the accuracy of the ray tracing.

source/RayTracing.cc
```diff
@@ -187,7 +225,19 @@ PointsValues<3> RayTracing::get_points_values()
...
   // different processors. Since the rays are on all the processors, we don't
-  // need to communicate the results to other processors.
+  // need to communicate the results to other processors. Note that we may lose
+  // some rays because the bounding boxes are larger than the cells and so a ray
+  // can be stopped by a bounding box and then miss the associated cell. We
+  // could get around this by asking ArborX for all the bounding boxes that the
+  // rays intersect but then we would need to keep track of the order of the
+  // intersections ourselves.
+  // TODO The current code can be simplified a lot once we can use ArborX 2.0
+  // This version of ArborX supports distribued ray tracing on triangle.
```
stvdwtt, line 235: Typo:
```suggestion
  // This version of ArborX supports distributed ray tracing on triangle.
```

source/RayTracing.cc
```diff
@@ -223,33 +273,51 @@ PointsValues<3> RayTracing::get_points_values()
...
+                                                  point_1[2] - point_0[2]});
+            dealii::Tensor<1, dim> const edge_02({point_2[0] - point_0[0],
+                                                  point_2[1] - point_0[1],
+                                                  point_2[2] - point_0[2]});
+            dealii::Tensor<1, dim> ray_direction(
+                {static_cast<double>(_rays_current_frame[i].direction[0]),
+                 static_cast<double>(_rays_current_frame[i].direction[1]),
+                 static_cast<double>(_rays_current_frame[i].direction[2])});
+            // double const direction_norm = ray_direction.norm();
+            // ray_direction /= direction_norm;
```
stvdwtt, line 320: Can these be deleted if they're not used anymore?

Rombur:
I have fixed the comments

stvdwtt (approved):

PR #377, opened 2025-03-11
Output solution just before and after data assimilation

Rombur:
With this PR, we output the solution just before and just after the data assimilation. This is on top of the "regular" outputs. This is on by default but it can be turned off to make a movie.

@stvdwtt I am not sure about the name of the input variable. If you have a better name, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,8 +163,8 @@ The following options are available:
 * memory\_space (optional): device (use GPU if Kokkos compiled with GPU support) or host (use CPU) (default value: host)
 * post\_processor (required):
   * filename\_prefix: prefix of output files (required)
-  * time\_steps\_between\_output: number of time steps between the
-  fields being written to the output files (default value: 1)
+  * time\_steps\_between\_output: number of time steps between the fields being written to the output files (default value: 1)
+  * output\_on\_data\_assimilation: output fields just before and just after data assimilation (default: true) **[since 1.1]**
   * additional\_output\_refinement: additional levels of refinement for the output (default: 0)
 * refinement (required):
   * n\_refinements: number of times the cells on the paths of the beams are refined (default value: 2)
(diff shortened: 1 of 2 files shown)


application/adamantine.hh
```diff
@@ -1717,6 +1718,7 @@ run_ensemble(MPI_Comm const &global_communicator,
       refinement_database.get("time_steps_between_refinement", 10);
   // PropertyTreeInput time_stepping.time_step
   double time_step = time_stepping_database.get<double>("time_step");
+  double const da_time_window = 2. * time_step;
```
stvdwtt, line 1721: Would it make sense to make this something like `2.1 * time_step` to make sure there's no round-off issue with the inequality check?

README.md
```diff
@@ -163,8 +163,8 @@ The following options are available:
 * memory\_space (optional): device (use GPU if Kokkos compiled with GPU support) or host (use CPU) (default value: host)
 * post\_processor (required):
   * filename\_prefix: prefix of output files (required)
-  * time\_steps\_between\_output: number of time steps between the
-  fields being written to the output files (default value: 1)
+  * time\_steps\_between\_output: number of time steps between the fields being written to the output files (default value: 1)
+  * output\_on\_data\_assimilation: output fields just before and just after data assimilation (default: true)
```
stvdwtt, line 167: Maybe `output_bracketing_data_assimilation`? Or `output_[on/upon/bracketing]_assimilation`? I'm also fine leaving it as-is.

application/adamantine.hh
Rombur: The name is a little bit misleading and I will change it. It's really half of the time window. I picked two to be sure to get the frame before and after the data assimilation. In practice, I tend to get three frames, one before the da and two after ir. We could use a sharper bound if we only want to see the data assimilation or larger one if we want to see some trend. I don't know what's better, so I picked two for now.

stvdwtt: Ah, I see that it's a half window now, thanks for pointing that out. Aesthetically, I'd prefer if the frames are symmetric. I don't know that we need to output more than two frames, so I think choosing the window size so that we always get one frame for the time step before assimilation and one where assimilation happens would be best.

PR #384, opened 2025-06-05
Output temperature gradients and cooling rates

Rombur:
This PR add a new option to output the gradient of the temperature, the cooling rate, and the interface velocity at the liquidus. The current format is `x y (z) temperature_gradient cooling_rate interface_velocity`. I can use `csv` if you prefer. Each processor writes its own data points to the file system and at the end of the simulation, rank zero concatenates all the files into a new one and removes the temporary ones. I thought it would be easier to read the data this way. Obviously this won't scale. 
Computing the quantities is pretty straightforward, the only pitfall is on the choice of the quadrature. By default, the quadrature points and the support points are different. So I changed the quadrature points to use the support points. This way we can compute the cooling rate directly using the temperature vector and the temperature gradient is computed at the right points. Otherwise, we would have to multiply the temperature vector with the shape function to evaluate the temperature at the quadrature points. Now we can skip that part and evaluate the gradient only if the dofs crossed the liquidus.


README.md
```diff
@@ -257,6 +257,8 @@ The following options are available:
       millimeter, centimeter, inch, or meter (default value: meter)
     * scan\_path: unit used for the scan path of the heat sources. Either
       millimeter, centimeter, inch, or meter (default value: meter)
+* microstructure (optional): **[since 1.1]**
```
stvdwtt, line 260: Consider to changing to something like `isotherm_metrics`, since we aren't directly calculating microstructure and that title gives us flexibility to add functionality to e.g. look at the martensite start temperature.

I guess one question is whether we intend to add more microstructure-specific functionality, like the JMAK model directly into adamantine. I had hoped to keep things simpler by doing JMAK as postprocessing, but for something like thermomechanics with in-situ precipitation, maybe we'd need to fully couple JMAK.

source/experimental_data_utils.cc
```diff
@@ -28,7 +28,8 @@ get_dof_to_support_mapping(dealii::DoFHandler<dim> const &dof_handler)
   std::unordered_set<dealii::types::global_dof_index> visited_dof_indices;
 
   // Manually do what dealii::DoFTools::map_dofs_to_support_points does, since
-  // that doesn't currently work with FE_Nothing
+  // that doesn't currently work with FE_Nothing. The issue has been fixed in
+  // deal.II 9.7
```
stvdwtt, line 32: As an aside, should we switch to the new deal.ii internal version?

tests/data/microstructure_G_R_gold_2.txt
```diff
@@ -0,0 +1,9 @@
+0 0 2 954.636364 477.318182
```
stvdwtt, line 1: Out of curiosity, did you verify the the gradient in this gold result is the expected one?

Rombur: yes, I did check.

README.md
Rombur: We decided to keep the name as-is for now. We will make a decision on the name once the JMAK work is more advanced.

PR #386, opened 2025-06-23
Fix a bug when averaging state ratios on a cell

Rombur:
This PR does three things:
- Fix a bug when averaging state ratios on a cell: we did not loop over all the quadrature points but only over the number of quadrature point in 1D. I realized this while working on a thermoelastic simulation. The displacement was not symmetric because we did not flag some cells as liquid.
- Add profiling sections in the mechanical component of the code
- Remove entries that are not read in an input file.


Rombur:
Retest this please

Issue #338, opened 2024-11-12
Radiative and convective BCs cool the part too quickly


stvdwtt:
We should do some more verification of the convective and radiative boundary conditions, they seem too strong. E.g. with just radiative conditions a part will cool from 1000K to 300K in 30 s. Similarly to get reasonable answers it looks like the convective coefficient needs to be on the order of 0.1, when references online suggest that 1-10 W/(m^2*K) is normal for free convection.

Ideally we'd find a good analytic test case, but we can also compare to e.g. AdditiveFOAM.

Rombur:
>  just to confirm, is the one in ThermalPhysics only used for GPU calculations and will be removed once the source calculation on the GPU is complete?

Yes, this only applies to the GPU calculations. See: 
https://github.com/adamantine-sim/adamantine/blob/d93ead353d087af286d6c7784abea0b70cbf5a94/source/ThermalPhysics.templates.hh#L87-L91

We don't compute the boundary conditions in `ThermalOperatorDevice`.

Rombur:
@AshGannon I think your working on this. If you are not feel free to unassign yourself.

AshGannon:
On it! Thank you for adding me to this issue.

AshGannon:
@Rombur you're right! Great catch.

PR #389, opened 2025-07-01
Fix a bug in non-adiabatic bc

Rombur:
This PR fixes https://github.com/adamantine-sim/adamantine/issues/338 The issue is that multiplied by an extra `fe_face_eval.get_value(q)` (i.e the temperature) the boundary value. This PR also disable non-adiabatic bc on the GPU. It needs more work. I've also added @AshGannon's  script to PR, so we can still do comparisons in the future. Below are the comparison between the analytical solution and the simulations, and the error.

analytical_vs_sim
error

It would be nice to also have an analytical solution for the radiative boundary condition but it is harder.


stvdwtt (approved):
This looks good to me! I'm headed out of town, so when the CI tests pass please feel free to merge yourself.

Rombur:
Retest this please

PR #390, opened 2025-07-03
Replace adamantine_COPY_INPUT_FILE with adamantine_COPY_DATA_FILE

Rombur:
This PR replace `adamantine_COPY_INPUT_FILE` with `adamantine_COPY_DATA_FILE`. The new function only has a single input. There is no need to specify where the file is located. I also reordered the file list alphabetically


Rombur:
Retest this please

PR #392, opened 2025-07-15
Add special path for forward Euler

Rombur:
This PR creates a special path when using forward Euler. When it takes less than one millisecond to perform a time step, going to deal.II and using a general Runge-Kutta method becomes noticeable.

The PR also adds Caliper marking around microstructure region.


AshGannon (approved):
Works on my end!

PR #393, opened 2025-07-17
Introduce Boundary class and 'clamped' bc

Rombur:
This PR reworks how we store boundary conditions and boundary ids. It consolidates everything into a new `Boundary` class. This PR also adds a new `clamped` boundary condition which imposes zero displacement in mechanical simulation. Before this PR, we always clamped the faces with a boundary id equals to four. Now we can clamp any boundary ids we want. Right now, user can set `clamped` bc with `boudary.type`. I am not sure if we should allow it or not because it won't do what you expect on the layer that's being printed. The top face won't be clamped but the thermal boundary condition will be applied. The current behavior is convenient but not consistent between the thermal and the mechanical parts of the simulation. What do you think @stvdwtt ?

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -1,10 +1,11 @@
-/* SPDX-FileCopyrightText: Copyright (c) 2016 - 2024, the adamantine authors.
+/* SPDX-FileCopyrightText: Copyright (c) 2016 - 2025, the adamantine authors.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
 #ifndef ADAMANTINE_HH
 #define ADAMANTINE_HH
 
+#include <Boundary.hh>
 #include <DataAssimilator.hh>
 #include <ExperimentalData.hh>
 #include <Geometry.hh>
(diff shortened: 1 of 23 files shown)


Rombur:
Retest this please

Rombur:
Retest this please

stvdwtt:
@Rombur, I really like the new `Boundary` class and the interface to it. I think it's good that the user can specify "clamped" for any boundary ID. In some way, it's the thermal BCs that are acting a little unexpected in that there's a thermal BC applied on a surface that doesn't exist on the full mesh.

A few comments on things that could be worth adding, but not necessarily in this PR:
1) We could add a special "printed_surface" boundary (either by boundary ID, is something like -1 an option) or otherwise, so the user can specify BCs on the activated/non-activated interface.
2) We could consider a run-time check on boundary conditions specified on boundaries that are fully deactivated at the start of the print. (Or maybe that would do more harm than good.)
3) I wonder if we should add `traction_free` as an option, as the counterpart of `adiabatic`. For thermal simulations there's no sense of "no BC", the user has to pick adiabatic. It seems to me that the same should be for mechanical BCs.

Rombur:
These are good points. I'll do them in a follow up PR

PR #398, opened 2025-08-19
Adding offset option for digital shadow work

AshGannon:
WIP, update this to a better method later on.


Rombur:
Closing this. @AshGannon has a temporary solution and we have a better solution long term

PR #399, opened 2025-08-20
Change format of experimental log file

Rombur:
I propose to change the way the build time log files work. Currently, we subtract the time of the first frame to all the following frames. This is very confusing and I don't know why we do that. @stvdwtt You wrote that code https://github.com/adamantine-sim/adamantine/pull/119 do you remember why we do this? 

Would this change break your workflow @stvdwtt @AshGannon ?

### DIFF
--- a/source/experimental_data_utils.cc
+++ b/source/experimental_data_utils.cc
@@ -162,8 +162,6 @@ read_frame_timestamps(boost::property_tree::ptree const &experiment_database)
   unsigned int num_cameras = last_camera_id - first_camera_id + 1;
   std::vector<std::vector<double>> time_stamps(num_cameras);
 
-  std::vector<double> first_frame_value(num_cameras);
-
   // Read and parse the file
   std::ifstream file;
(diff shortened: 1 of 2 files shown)


stvdwtt:
I think the motivation is that we wanted to account for the case where different "cameras" have times that are shifted relative to each other. Subtracting off the first frame time fixes that shift -- but with the weird assumption that these un-synced cameras would emit their first frame at the same real time.

For now I'm fine to just assume that the cameras are synced -- if both data sources come from ROS that should definitely be true. For robustness in the future we might want to add per camera time offsets.

Rombur:
>  In that case maybe the observations were accidently shifted by 30 s, skipping the first one?

Yes, that's what I think too. I was trying to do data assimilation with the data that @AshGannon gave me and instead of simply shifting everything by 30 s, I used the exact timings from ROS. That gave me unexpected results and that's why I looked at the code.

Rombur:
@AshGannon what do you think of this change?

PR #403, opened 2025-09-16
Remove implicit time-stepping methods

Rombur:
I don't think this really works unless you have adiabatic boundary conditions and the material properties are independent of the temperature. So let's remove that code. If we need implicit methods, we should re-implement them cleanly.


Rombur:
@stvdwtt are you OK with removing the implicit time-stepping

stvdwtt:
I'm pretty torn on this point. While we don't use implicit methods now and it would take some time/effort to get implicit methods fully working (e.g. improving the nonlinear solve), I think there are cases where implicit could give huge speedups -- like in the Okuma prints we've been doing where we spend the vast majority of the time dwelling.

Rombur:
Implicit methods would very useful and I am not suggesting that we will never support them. However the current implementation does not work and I don't see us making any progress in the next 12 months. This is essentially dead code right now, and I don't want to give the false impression that it works. 

If you really want to keep the code, I'll scrub the documentation and the input files to remove any mention of  implicit methods.

PR #427, opened 2025-12-29
heat source integration

colemanjs:
Added the the volume integration of the total heat input to the system for monitoring and verification.

### DIFF
--- a/source/ThermalPhysics.hh
+++ b/source/ThermalPhysics.hh
@@ -192,6 +192,10 @@ private:
    */
   std::vector<std::shared_ptr<HeatSource<dim>>> _heat_sources;
   /**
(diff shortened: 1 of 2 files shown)


Rombur:
We should not do this unconditionally. Either guard the new code with `#if ADAMANTINE_DEBUG` so that it only runs in debug mode or have an `if` condition that can be controlled from the input file.

colemanjs:
> guard the new code with `#if ADAMANTINE_DEBUG` so that it only runs in debug mode or have an `if` condition that can be controlled from the input file.

Added the optional flag `print_heat_input` to the `sources` dictionary in ac042a8

Rombur:
Looks good. The CI failed because of an issue with the machine. I've restarted it. I'll merge the PR once the CI is done.

Issue #433, opened 2026-01-12
Frontier test failures


masterleinad:
On Frontier, I'm seeing with ROCm 6.2 (default)
```
	  1 - indent_code (Failed)
	  2 - test_integration_da_1 (Subprocess aborted)
	  3 - test_integration_da_2 (Subprocess aborted)
	  4 - test_integration_da_3 (Subprocess aborted)
	  5 - test_integration_da_6 (Subprocess aborted)
	  7 - test_experimental_data_1 (Timeout)
	 10 - test_integration_3d_device_1 (Subprocess aborted)
	 11 - test_material_property_1 (Subprocess aborted)
	 13 - test_mechanical_operator_1 (Subprocess aborted)
	 14 - test_mechanical_physics_1 (Subprocess aborted)
	 15 - test_post_processor_1 (Subprocess aborted)
	 18 - test_thermal_operator_1 (Subprocess aborted)
	 19 - test_thermal_operator_device_1 (Subprocess aborted)
	 24 - test_integration_2d_1 (Subprocess aborted)
	 25 - test_integration_2d_2 (Subprocess aborted)
	 28 - test_integration_3d_1 (Timeout)
	 29 - test_integration_3d_2 (Timeout)
	 30 - test_integration_3d_amr_1 (Subprocess aborted)
	 31 - test_integration_3d_amr_2 (Subprocess aborted)
	 34 - test_integration_da_augmented_1 (Subprocess aborted)
	 35 - test_integration_da_augmented_2 (Subprocess aborted)
	 36 - test_integration_thermoelastic_1 (Timeout)
	 37 - test_integration_thermoelastic_2 (Timeout)
	 38 - test_material_deposition_1 (Subprocess aborted)
	 39 - test_material_deposition_2 (Subprocess aborted)
	 40 - test_microstructure_1 (Subprocess aborted)
	 41 - test_microstructure_2 (Subprocess aborted)
	 42 - test_thermal_physics_1 (Subprocess aborted)
	 43 - test_thermal_physics_2 (Subprocess aborted)
```
where most of these come from something like
```
5: Test command: /lustre/orion/ums018/proj-shared/adamantine/build/bin/test_integration_da
5: Working Directory: /lustre/orion/ums018/proj-shared/adamantine/build/tests
5: Test timeout computed to be: 1500
5: Running 3 test cases...
[...]

PR #437, opened 2026-01-14
Ignore FE_INVALID in test_experimental_data for ArborX calls

masterleinad:
Fixes part of #433. Acoording to https://github.com/arborx/ArborX/blob/71f0c9664372631bcd5cb124a0815de1f02528b5/src/geometry/ArborX_Ray.hpp#L84-L157, the box-ray intersection algorithm relies on NaN behavior which we trigger in `test_experimental_data`. Thus, this pull request disables floating point expections for the problematic ArborX calls.

### DIFF
--- a/tests/test_experimental_data.cc
+++ b/tests/test_experimental_data.cc
@@ -168,14 +168,16 @@ BOOST_AUTO_TEST_CASE(read_experimental_data_ray_tracing_from_file)
   ray_tracing.read_next_frame();
 
   // Since the ArborX implementation relies on NaN comparisons here, we need to
-  // allow FE_INVALID.
+  // allow FE_INVALID. This should be resolved in ArborX 2.0.
 #ifndef __APPLE__
   std::fenv_t envp;
   std::feholdexcept(&envp);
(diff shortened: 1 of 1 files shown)


masterleinad:
> Which version of ArborX are you using?

That's with ArborX 1.7. https://github.com/arborx/ArborX/pull/1226 is only part of ArborX 2.0.

tests/test_experimental_data.cc
```diff
@@ -167,8 +167,15 @@ BOOST_AUTO_TEST_CASE(read_experimental_data_ray_tracing_from_file)
...
 
+  // Since the ArborX implementation relies on NaN comparisons here, we need to
+  // allow FE_INVALID.
+#ifndef __APPLE__
+  std::fenv_t envp;
+  std::feholdexcept(&envp);
+#endif
   // Compute the intersection points
   auto points_values = ray_tracing.get_points_values();
+  std::fesetenv(&envp);
```
Rombur: You need to guard with `#ifndef __APPLE__`

PR #456, opened 2026-02-16
Only assemble the mechanical system matrix when necessary

Rombur:
This PR split `assemble_system` in two: `assemble_matrix` and `assemble_rhs`. `assemble_rhs` is called every time steps while `assemble_matrix` is only called when the mesh or the dof indices are changed.

Alternative to https://github.com/adamantine-sim/adamantine/pull/444

### DIFF
--- a/application/adamantine.hh
+++ b/application/adamantine.hh
@@ -1,4 +1,4 @@
-/* SPDX-FileCopyrightText: Copyright (c) 2016 - 2025, the adamantine authors.
+/* SPDX-FileCopyrightText: Copyright (c) 2016 - 2026, the adamantine authors.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  */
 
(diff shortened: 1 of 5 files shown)


Rombur:
@masterleinad We messed up. The matrix actually changes all the time because we only compute the mechanics on the solid part not on the liquid. This changes even if we don't touch the mesh. We update the fe indices in `setup_dofs` to filter out the melted cells.

masterleinad:
> @masterleinad We messed up. The matrix actually changes all the time because we only compute the mechanics on the solid part not on the liquid. This changes even if we don't touch the mesh. We update the fe indices in `setup_dofs` to filter out the melted cells.

https://github.com/adamantine-sim/adamantine/pull/444 takes that into account AFAICT.

Rombur:
That's right. 

We should still review the assembly and probably use multithreading for it too. You saw a big performance improvement because it takes time for the material to melt. If you just look at the first few seconds, you can reuse the sparse matrix. Once you start printing, you need to re-assemble the sparse matrix almost all the time, especially if you use the `mechanical_time_step_factor` (unless you dwell).

PR #472, opened 2026-04-27
Use three cores when runing test_integration_3d in debug mode

Rombur:
We have a couple of PRs that are timing out on `test_integration_3d`. The issue is that this test runs very slowly in debug mode. On some machines it completes on time, but on others it times out. This PR uses 3 cores to run the test in debug.


Rombur:
Actually let's not do that. I thought that we were testing both Debug and Release versions of the code but we only test Debug. Since the problem  only appears for the CPU configuration, I will just switch this configuration to Release. That way we test both configurations.

PR #473, opened 2026-04-27
CI: run the CPU configuration in Release mode

Rombur:
Supersede #472

### DIFF
--- a/source/PostProcessor.cc
+++ b/source/PostProcessor.cc
@@ -103,30 +103,34 @@ dealii::Vector<double> PostProcessor<dim>::get_von_mises_stress(
   // currently just average the values on the cells.
   dealii::Vector<double> von_mises_stress(
       _mechanical_dof_handler->get_triangulation().n_active_cells());
-  unsigned int const n_quad_pts =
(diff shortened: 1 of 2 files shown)


source/PostProcessor.cc
```diff
@@ -103,30 +103,34 @@ dealii::Vector<double> PostProcessor<dim>::get_von_mises_stress(
   // currently just average the values on the cells.
   dealii::Vector<double> von_mises_stress(
       _mechanical_dof_handler->get_triangulation().n_active_cells());
-  unsigned int const n_quad_pts =
-      stress_tensor.size() > 0 ? stress_tensor[0].size() : 0;
-  unsigned int cell_id = 0;
-  for (auto const &cell : _mechanical_dof_handler->active_cell_iterators())
+  // Currently we only support computing the von mises in 3D.
+  if constexpr (dim == 3)
```
masterleinad, line 107: That looks unrelated.

Rombur: We are accessing the element `[2][2`] in a `Tensor<2,dim>` which does not work if `dim = 2`. I don't know why we don't get an error in debug.

PR #474, opened 2026-04-28
CI: use ctest -j4

Rombur:
Let's see if this works as expected.


Rombur:
This PR makes the `test_integration_3d` tests time out. We should revisit these tests and understand why they are so slow in debug.

PR #484, opened 2026-07-01
Update the Adamantine logo

stvdwtt:
This now uses the the MDF-generated Adamantine logo. It does not update the QR code. If the documentation site points to these files, we may have to make edits there too.


Rombur:
> If the documentation site points to these files, we may have to make edits there too.

It does. Any reason why you didn't overwrite the file instead of creating new ones?

stvdwtt:
No good reason, let me fix that. I pushed a commit with filename changes, but I'll push another one that adds white background versions and reverts the README. I haven't found an easy way to update the QR code, but I'll try a couple more things.

stvdwtt:
Reverted to previous file names and all files are now updated (including the QR code).

Rombur (approved):

Rombur:
Looks good. Thanks.

I merging without waiting for the CI since the code is unchanged

PR #486, opened 2026-07-14
Jenkins CI: do not pull Docker image every time because we reach the rate limit

Rombur:
The CI does not pass anymore because we reach the rate limit. Since we update the image once or twice a year, I'll just go on the different machines and update the images manually.

Second change: according to copilot the failure on the nix macos build is due to a problem upstream. I changed the CI, so that nix ubuntu runs even if nix macos fails

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -6,6 +6,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
+    continue-on-error: ${{ matrix.os == 'macos-latest' }}
     runs-on: ${{ matrix.os }}
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


Rombur:
@codex please review this PR

Rombur (commented):
Reviewed by ChatGPT on behalf of Bruno Turcksin.

I reviewed the proposed change. The modification is straightforward: switching `alwaysPull` from `true` to `false` for both Jenkins Docker agents avoids repeated image pulls and addresses Docker Hub rate limiting while relying on manually refreshed images.

I did not identify any correctness issues with the change itself. The only suggestion is to document the operational assumption somewhere (either in `ci/jenkins_config` or CI documentation): the Docker images on Jenkins workers must be updated manually whenever `rombur/adamantine-stack:*` changes. Without that documentation, future maintainers may be surprised that CI is not automatically picking up newer container images.

Otherwise, this looks reasonable to me.

Rombur:
@codex please review this PR

PR #491, opened 2026-09-15
Add a thermoelastoplastic test

Rombur:
A fully clamped thermoelastoplastic body with an injected deviatoric prior stress. It checks that the hydrostatic thermal stress does not change the J2 norm or flow direction, while plastic and thermal history are updated over two increments.


Rombur (commented):
Codex/ChatGPT review — posted at the user's request. This review is written by Codex, not by the user or the PR author.

I found no actionable correctness issues in the added test. The expected hydrostatic thermal stress is consistent with a fully clamped isotropic body, and the second collinear increment exercises the stored plastic and kinematic history through the resulting stress. The PR's Jenkins check was still pending at the time of review.

PR #494, opened 2026-09-18
Add activation_visualizer to visualize element activation

Rombur:
This PR adds a new executable `activation_visualizer` that shows which and when the cells will be activated.  We take advantage that `Adamantine` is composed of a library and an executable to reuse a lot the existing code. Cells that haven't been activated yet have time stamp of `-1`, so that they can be filtered out.


applications/activation_visualizer.cc
```diff
@@ -0,0 +1,314 @@
...
+    boost::property_tree::ptree const &beam_database =
+        sources_database.get_child("beam_" + std::to_string(beam_id));
+    std::string const scan_path_file =
+        beam_database.get<std::string>("scan_path_file");
+    std::string const scan_path_file_format =
+        beam_database.get<std::string>("scan_path_file_format");
+
+    adamantine::ScanPath scan_path(scan_path_file, scan_path_file_format,
+                                   units_optional_database);
+    deposition_paths.push_back(adamantine::deposition_along_scan_path<dim>(
```
Rombur: [P1] Honor the configured material-deposition mode. This helper always reconstructs deposition from `sources.beam_i.scan_path_file`, so an input with `geometry.material_deposition` disabled still produces activation events, while `material_deposition_method = file` ignores `geometry.material_deposition_file` (and may fail because scan-path/deposition-size keys are absent). The main application delegates this dispatch to `create_material_deposition_boxes`, which returns no boxes when deposition is disabled and uses `read_material_deposition` for file mode. Please use the same dispatch here.

applications/activation_visualizer.cc
```diff
@@ -0,0 +1,314 @@
...
+  unsigned int output_time_step = 0;
+  write_activation_output(communicator, triangulation, output_directory,
+                          filename_prefix, output_time_step, 0.,
+                          activation_time, times_filenames);
+  ++output_time_step;
+
+  // The material-deposition API returns one cell list for every deposition
+  // box. Apply those lists in chronological order and write one cumulative
+  // snapshot whenever the deposition time changes.
+  for (unsigned int i = 0; i < elements_to_activate.size(); ++i)
```
Rombur: [P2] Group boxes with the same deposition timestamp before writing output. `merge_deposition_paths` sorts boxes but preserves equal times, and this loop writes a PVTU after every box. With coincident boxes (for example, multiple beams), the PVD contains duplicate timestamps and the first snapshot contains only a subset of the cells that should activate at that time. Accumulate all consecutive equal-time boxes, then write one cumulative snapshot.

Rombur (commented):
Codex/ChatGPT review — posted at the user's request. This review is written by Codex, not by the user or the PR author.

I found two functional issues in the activation visualizer; both are called out inline below. The Jenkins PR check for this commit is currently successful.

